PR #12, opened 2022-05-18
Stronger tests, and a couple optimizations in circular arrays.

fpottier:
This PR proposes:
* A reasonably strong sequential test: `make -C seqtest random`. (Requires `monolith`.)
* A reasonably strong concurrent test: `dune runtest --force`.
* An optimization in circular arrays: remove the redundant `mask` field and remove an indirection.
* An optimization in circular arrays: use `Array.blit` instead of a loop when growing or shrinking.

### DIFF
--- a/seqtest/.gitignore
+++ b/seqtest/.gitignore
@@ -1,0 +1,3 @@
+input
+output
+dune-workspace.afl
--- a/seqtest/Makefile
+++ b/seqtest/Makefile
@@ -1,0 +1,3 @@
+SWITCH := 5.0.0+trunk
+EXE := seqtest.exe
+include $(shell ocamlfind query monolith)/Makefile.monolith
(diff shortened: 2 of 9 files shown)


fpottier:
Apparently the fact that the code in `seqtest/` requires `monolith` causes continuous integration to fail. I will see if I can make some changes so that this code is not built unless explicitly requested by the user.

kayceesrk (approved):
@fpottier Thanks for the stronger tests and the simplification. The changes look good to me. 

I came to know of `monolith` thanks to this PR. I thought I would ping @jmid to see whether he's interested in having a look at this PR as he's building similar tools for concurrency testing.

fpottier:
Hi @kayceesrk! I may have further suggestions for improvements in the future. Regarding `monolith`, it is indeed limited to sequential testing scenarios at the moment. I have not yet thought about generalizing it to a concurrent setting; that would definitely require a more complex description language, as one would need to explain exactly how a concurrent object is intended to be used.

PR #13, opened 2022-05-30
Tests for Ws_deque using QCheck

lyrm:
Adds tests for `ws_deque` data structure for different domain configuration :
- a producer alone
- a producer and a single stealer
- a producer and two stealers.


Sudha247 (approved):
Thanks very much for the tests! (apologies for the delay in getting to this). I've read through the tests and they do what's intended. I'm going to try and fix the CI before merging.

PR #16, opened 2022-06-22
Add multiple-producer single-consumer queue from eio repo

lyrm:
This PR adds the  multiple producers, single consumer queue named `lf_queue` in [eio](https://github.com/ocaml-multicore/eio). It has been renamed `mpsc_queue` and `qcheck` tests has been added. 

`qcheck` tests checks : 
- some simple properties that are closed to the functions semantics with only one domain (producer or consumer)
- multiple domains interaction with two configurations (one producer + one consumer) and (two producers + one consumer)

A PR for more tests will be done in the [multicoretests](https://github.com/jmid/multicoretests) repository as soon as this PR is merged.


src/mpsc_queue.mli
```diff
@@ -0,0 +1,32 @@
+(** A lock-free multi-producer, single-consumer, thread-safe queue without support for cancellation.
+    This makes a good data structure for a scheduler's run queue. *)
```
Sudha247, line 2: Minor nit: it'd be useful to clarify we're talking about eio's scheduler here.

lyrm:
Sorry for the forced push, I try multiple things but none weren't resulting in a messy log.

PR #21, opened 2022-10-07
MPMC relaxed queue

bartoszmodelski:
MPMC Queue

This PR adds: 
- FAD MPMC Queue - pretty modern FAD queue. Described in detail in `mpmc_queue.ml`. 
- CAS MPMC Queue (see CAS_interface in `mpmc_queue.ml`). Standard CAS queue, which may perform better on architectures without fetch and add. On x86 its much worse, e.g. 200% slower in the balanced read-write case. 
- Tests 
- Benchmarks, including a shell script to run a few configurations and dump text file. I think this output should live in the repo to communicate what to expect, and track performance changes. Nonetheless it is a very minimal setup atm., so any feedback is super welcome. We could build a jupyter notebook to run all benchmarks and show some charts (there's some interesting stuff going on!), but I don't wanna jump too far forward.


bench/stats.ml
```diff
@@ -0,0 +1,28 @@
...
+        curr_sum +. squared_diff)
+      0. data
+  in
+  let n = Int.to_float (List.length data) in
+  Float.sqrt (sum /. n)
+
+let cmp ?(epsilon=0.01)  a b = 
+  Float.abs (a -. b) < epsilon;;
+
+let sanity_checks () =
```
Sudha247, line 21: Useful to have this for sure. Would it be suitable to put this in a test file though?

Sudha247 (commented):
Thanks for the PR! It's always nice to have more lockfree structures. :)

I haven't done a full review, some minor questions/comments below.

test/test_mpmc_queue.ml
```diff
@@ -0,0 +1,112 @@
...
+  let tail_val = Atomic.get tail in
+  Alcotest.(check int) "hd an tl match" head_val tail_val;
+  Array.iter
+    (fun item ->
+      Alcotest.(check (option int))
+        "ghost item in the queue!" None (Atomic.get item))
+    array
+
+let () =
+  let open Alcotest in
```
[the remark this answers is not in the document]
bartoszmodelski: Fwiw, under the hood this is not a new dependency. Quickcheck tests are actually converted and executed by Alcotest (https://github.com/ocaml-multicore/lockfree/blob/e1396b0a038cdc32cf0c9087548a49d3e9c9c934/test/qcheck_ws_deque.ml#L274). I don't want to use Quickcheck, but still wrapped my tests in the same Alcotest "harness" to make the test output look nice and consistent with others.

bench/stats.ml
bartoszmodelski: True. Since these are not the usual lock free structure tests, do you think it is more customary to put these in `/test/test_stats.ml` or just `/bench/test_stats.ml`?

src/mpmc_queue.mli
```diff
@@ -0,0 +1,28 @@
...
+    
+    Note: in extreme cases the structure may occassionaly violate FIFO 
+    order. 
+*)
+
+type 'a t = private {
+  array : 'a Option.t Atomic.t Array.t;
+  head : int Atomic.t;
+  tail : int Atomic.t;
+  mask : int;
```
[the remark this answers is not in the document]
bartoszmodelski: I'm asserting on these in the tests. I guess it could be made a little cleaner in the following way: 

```
type 'a t 

... 

module Testing : sig 
  val assert_empty_invariants : unit -> unit 
end 
```

lyrm:
Hi! Thanks for the PR. I will be reviewing it this week.

lyrm:
I have written my own `dscheck` tests for this queue, and I have an issue with this test where two domains try to push on a full queue. It seems to be stuck (looping ?). 
 
````ocaml
let _test () =
  let t= Mpmc_queue.create ~size_exponent:1 () in
  let t1 = Mpmc_queue.push t 1 in
  let t2 = Mpmc_queue.push t 2 in
  assert (t1 && t2);

  let is_pushed1 = ref false in
  let is_pushed2 = ref false in

  Atomic.spawn (fun () -> is_pushed1 := Mpmc_queue.push t 0);
  Atomic.spawn (fun () -> is_pushed2 := Mpmc_queue.push t 0);
  Atomic.final (fun () ->
      total_checked := !total_checked + 1;
      Atomic.check (fun () -> (not !is_pushed1) && not !is_pushed2))
 ````

I am going to  try to reproduce it outside `dscheck`. Any idea why it is not working properly ?

**Comment by bartoszmodelski** [comment]
I think so. 

Background: both push and pop are optimistic in that they modify head/tail first, and check the queue afterwards. If queue is full or empty, push/pop have to somehow fix the state before returning with failure. It is done by rolling back the previously-incremented index, or moving forward the other one. These operations must have exactly the reverse effect of initial fetch-and-add, and are done with CAS. Now, if the queue is empty and there's a few pushers, they will form a chain. Only the first or the last can free themselves. The first one by moving head, the last one by rolling back tail. Normally, this will happen smoothly, and, progressively, everyone in the chain is going to free itself. 

However, if there's more pushers (or poppers) than cells, there can be a pathological situation. Assume 4 pushers on a 2-cell array. The first one gets index 0 (slot 0), the fourth one index 3 (slot 1). They both successfully insert items, and now pushers that got indices 1, 2, are unable to complete any rollback operation. 

In other words, normally, the chain of pushers (or poppers) is confined to one end of queue. But if the array is too small, it may occur inside it. 

I think we could prevent this behavior by enforcing FIFO on overlap as is done in [here](https://github.com/rigtorp/MPMCQueue).

**Comment by lyrm** [comment]
>  These operations must have exactly the reverse effect of initial fetch-and-add, and are done with CAS. Now, if the queue is empty and there's a few pushers, they will form a chain. Only the first or the last can free themselves. The first one by moving head, the last one by rolling back tail. Normally, this will happen smoothly, and, progressively, everyone in the chain is going to free itself.

(I'm guessing you meant if there are a few poppers and the queue is empty)

What you describe seems to be a blocking situation:
- create a new queue of size >> number of domains (=3 in our case)
- 3 domains simultaneous try to pop : 
    - domain 1 got head = 0
    - domain 2 got head = 1
    - domain 3 got head = 2
    - global head = 3 and global tail = 0 
- domain 1 and 3 get suspended just after their `fetch_and_add` operation on `head`
- domain 2 is blocked as neither `cas` will work.

**Comment by lyrm** [comment]
> However, if there's more pushers (or poppers) than cells, 

Unfortunately, the test I wrote get the same issue with a bigger queue and 2 domains. It seems it just needs to be full.

**Comment by bartoszmodelski** [comment]
> (I'm guessing you meant if there are a few poppers and the queue is empty)
> 
> What you describe seems to be a blocking situation:
> 
> * create a new queue of size >> number of domains (=3 in our case)
> * 3 domains simultaneous try to pop :
>   
>   * domain 1 got head = 0
>   * domain 2 got head = 1
>   * domain 3 got head = 2
>   * global head = 3 and global tail = 0
> * domain 1 and 3 get suspended just after their `fetch_and_add` operation on `head`
> * domain 2 is blocked as neither `cas` will work.

Agreed, this case will not terminate in DSCheck. I would argue it's a non-issue as the actual time spent blocked is going to be small, especially in a busy queue. 

I was referring to a different case, described below, which seems more nefarious to me. 

```
Assume 4 domains on a 2-cell queue, each tries to push. 

* domain 1 got tail = 0
* domain 2 got tail = 1 
* domain 3 got tail = 2 (ind 0)
* domain 4 got tail = 3 (ind 1)
```

Domains 2, 3 are suspended right after fetch_and_add on tail. Now, domains 1, 4 succeed at placing their items and terminate. That leaves domains 2, 3, unable to insert elements (ind 0, ind 1 are occupied), unable to roll-back (tail is too far) and unable to move forward (head is too low). This leaves them indefinitely blocked. This is bad, but I think it requires number of domains to be higher than size of array.

**Comment by bartoszmodelski** [comment]
Rebased on top of main, updated benchmark setup and did the rebase trick to remove lots of stale diff.

File: src/mpmc_queue.ml [hunk 4]
```diff
@@ -0,0 +1,245 @@
...
+      (* succedded to push *)
+      true
+    else if ccas tail (tail_val + 1) tail_val then (* rolled back tail *)
+      false
+    else if
+      time_to_try_push_forward nth_attempt && ccas head tail_val (tail_val + 1)
+    then (* pushed forward head *)
+      false
+    else (* retry *)
+      take_or_rollback (nth_attempt + 1)
```
lyrm, line 152: This function can prevent a thread from progressing if other threads are paused which breaks the definition of obstruction-free (and thus is not lock-free).

src/mpmc_queue.ml
```diff
@@ -0,0 +1,245 @@
...
+        (* dequeued an item, return it *)
+        value
+      else if ccas head (old_head + 1) old_head then (* rolled back head *)
+        None
+      else if
+        time_to_try_push_forward nth_attempt && ccas tail old_head (old_head + 1)
+      then (* pushed tail forward *)
+        None
+      else take_or_rollback (nth_attempt + 1)
+    in
```
lyrm, line 193: Same comment as for `push`. If a domain holds a _middle_ value for `old_head`, meaning `old_head` is neither equal to `tail` nor to `head-1` and the queue is empty, none of the 3 possible CAS of this function can work. The only way for the domain to progress is for _other_ domains to change the value of either `head` or `tail`. This is thus not obstruction-free.

It is possible for a domain to hold such a _middle_ value for `old_head` if 3 domains tries to pop simultaneously from an empty queue.

lyrm (commented):
As I discussed with @bartoszmodelski, there is an issue in this implementation with how a thread can progress through the `take_or_rollback` function. The two concerned parts are not obstruction-free as defined in _On the nature of Progress_ (Herlihy and Shavit).

Here is the definition of obstruction-free from the paper: 

> We say that a method call executes in isolation for a duration if no other threads take steps during that time. A method is obstruction-free if every thread that calls that method returns if that thread executes in isolation for long enough. This condition is non-blocking, and is strictly weaker than the lock-free condition.

lyrm:
Also, as discussed (and  as mentioned in its comments), this implementation does not always preserve FIFO order and thus should be rename accordingly. Your suggestion to add `relaxed` in the name  would be great. 
Thanks !

bartoszmodelski:
Thanks for the review!

I'll rename the queue accordingly and let's fix the ordering in a separate PR. 

Regarding the lock-free property: should we go forward with the original interface proposed in the paper, and perhaps add above methods as non-lockfree, since they are often more useful?

lyrm:
Thanks for these new commits ! The current interface looks good. I will finish reviewing this PR at the beginning of next week. I need to look at everything else than implementation!

src/mpmc_relaxed_queue.ml
```diff
@@ -0,0 +1,269 @@
...
+
+  It's present in the literature. Weakly I'm not a fan. If dequeuers are faster to remove 
+  items than enqueuers supply them, slots burned by dequeuers are going to make enqueuers 
+  do even more work.
+
+  # Resizing 
+
+  The queue does not support resizing, but it can be simulated by wrapping it in a 
+  lockfree list. 
+*)
```
lyrm, line 96: Is this documentation up to date with the last changes ?

lyrm (commented):
### About documentation 
Some `(*` to replace with `(**`

### About bench 

The bench repository would need a `README` with a short note about how to run them, what they do and how to read them (if it makes sens).

src/mpmc_relaxed_queue.ml
bartoszmodelski: Yeah. Added a couple corrections just now to refer to non-lf paths.

bartoszmodelski:
> ### About documentation
> Some `(*` to replace with `(**`
> 
> ### About bench
> The bench repository would need a `README` with a short note about how to run them, what they do and how to read them (if it makes sens).

Addressed both

PR #24, opened 2022-10-20
Ws deque benchmark

bartoszmodelski:
Based on #23

Benchmark for structures implementing ws_deque, which is the interface accepted by Domainslib. Currently, we have two structures: 
- SPMC Queue
- Deque


lyrm:
Done later on with `multicore-bench.`

PR #25, opened 2022-10-20
Add a comment about deque 

bartoszmodelski:


lyrm:
I have squashed the commit history and added the reference to the papers used  for the implementation in the README. 

I will merge this PR as soon as the checks are done.

Sudha247:
The CI failure is just the benchmark run failure. It's being investigated by @bartoszmodelski in #28. That need not block this PR.

PR #27, opened 2022-10-27
Add Michael Scott queue and Treiber stack

tmcgilchrist:
Add Michael Scott queue and Treiber stack using `Atomic` from OCaml 5 and re-introduces the `Backoff` module using `Domain.cpu_relax ()`. 

Outstanding questions for me:
-  Names for the modules, I am not sure what names to put them under? 
- More tests to confirm they work as expected. I am still writing those.
-  Does this need benchmarks? I wasn't planning on adding them but I can with some guidance on how to write/run them
- Is the Backoff module required?


kayceesrk:
@lyrm @bartoszmodelski How do you see this PR fit within the rest of the developments in this repo? IINM there is no implementation of these modules already in development, in which case, I would be interested in finding out what needs to be done to get these data structures into the repo.

bartoszmodelski:
Thanks for the PR! 

Both structures are useful to have. We don't have any mpmc stacks or resizable mpmc queues yet (or on the way). I think we need to polish the PR a little bit and it will be good to merge. Happy to help here. 

* Names for the modules, I am not sure what names to put them under?
Maybe just TreiberStack and MichaelScottQueue? 

* More tests to confirm they work as expected. I am still writing those.
Sounds good. On top of the usual tests, [dscheck](https://github.com/sadiqj/dscheck) is a good tool for validation of edge cases. The end goal is to integrate dscheck with lockfree repo and run it for everything, so if you decided to take this path, these tests would get merged in some (hopefully) not so distant future as well.  

* Does this need benchmarks? I wasn't planning on adding them but I can with some guidance on how to write/run them
Yes, we aim to have something for every structure. We're in the process of setting up continuous benchmarking now, so just something minimal will do. [Queue example](https://github.com/ocaml-multicore/lockfree/pull/21/files#diff-f2e8dd51892d6dbc663b791b1f8561d45b2ab42fa83897cbbad66c0f70cb6e07)

* Is the Backoff module required?
Kinda looks like it was done to optimize an undercommited workload, and could perform poorly on an overcommited one. It is fairly surprising nonetheless. 

My thinking would be to get rid of it, unless we have good evidence for keep it. 



Other than that, we should also add some docs with some minimal context. I would just state the linearizations (multi-producer multi-consumer I think?) and highlight its dynamic size. Perhaps we should also warn that structure may yield. [This is what I wrote for deque.](https://github.com/ocaml-multicore/lockfree/pull/25/files)

lyrm:
Thanks for this PR. I will have a quick look at it next week but will need more time to make a proper review. 

As @bartoszmodelski said, we don't have these specific data structures here yet,  so this is a welcomed addition. I will also add that, as it is inspired by well known papers and algorithms, it will be easier to get some guaranties on the properties of these Ocaml implementations but also to determine their possible known strength and drawbacks.

src/lf_stack.ml
```diff
@@ -0,0 +1,62 @@
+(** Trieber's Lock Free stack *)
+
+type 'a node =
+  | Nil
+  | Next of 'a * 'a node Atomic.t
+
+type 'a t =
+  { head : 'a node Atomic.t ;
+    tail : 'a node Atomic.t }
```
bartoszmodelski, line 9: tail is not used, perhaps a leftover from this being implemented on the base of msqueue?

src/lf_stack.ml
```diff
@@ -0,0 +1,62 @@
...
+
+let pop q =
+  let rec loop b () =
+    let s = Atomic.get q.head in
+    let newhead = match s with
+      | Nil -> failwith "Stack.pop: impossible"
+      | Next (_, x) -> Atomic.get x
+    in match newhead with
+       | Nil -> None
+       | Next (v, _) when Atomic.compare_and_set q.head s newhead -> Some v
```
bartoszmodelski, line 48: Afaict this has ABA problem. If we read `newhead`, and in the meantime:
1. Other thread pops it
2. Other thread insert a new node, which won the lottery and got the same reference.
This CAS will succeed despite looking at a different node, and function will return a wrong value (old one). 

In the original implementation/paper this is solved by including a counter in the CAS. 

While theoretically possible, in our case gc would have to run between points 1., 2. to make that scenario possible. I wonder if there is any way to guarantee that that does not happen, e.g. disable safepoints or anything.

src/backoff.ml
```diff
@@ -0,0 +1,17 @@
...
+
+let create ?(max = 32 * 4096) () = (max, ref 1)
+
+let once (maxv, r) =
+  let t = Random.State.int (Domain.DLS.get k) !r in
+  r := min (2 * !r) maxv;
+  if t = 0 then ()
+  else begin
+      for _ = 1 to t do
+        Domain.cpu_relax ()
```
bartoszmodelski, line 13: Backoff module looks good to me. I thought the implementation of relax is different, but it's doing what's desired (pause). Feel free to mark as resolved.

src/lf_stack.ml
bartoszmodelski, line 9: tail not used, maybe a leftover from this being implemented from msqueue?

src/backoff.ml
bartoszmodelski: perhaps it's worth adding an `.mli` and stating that Backoff busy-waits but in a way that's nicer to ocaml runtime and system than a simple `while .. do () done`

src/lf_msqueue.ml
```diff
@@ -0,0 +1,100 @@
...
+  let newnode = Next (v, Atomic.make Nil) in
+  let tail = Atomic.get q.tail in
+  match tail with
+  | Nil         -> failwith "HW_MSQueue.push: impossible"
+  | Next (_, n) -> begin
+      find_tail_and_enq n newnode;
+      ignore (Atomic.compare_and_set q.tail tail newnode)
+  end
+
+let clean_until q f =
```
bartoszmodelski, line 72: I'm suspicious that this function may not linearize if called from more than 1 thread. I'll try to create a test that breaks it. 

Assume queue with the following values [1; 2; 1; 2; 1; 2; ...]. If we call [clean_until] from two threads: 
* A wants to delete until value is 1 
* B want to delete until value is 2

We'd expect at most 3 items to be removed, depending whether thread A or B linearize first. But it seems that they could delete more than that if thread A happens to be running when 2 is the value of `head`, and thread B happens to be running when `1` is the `head`.

bartoszmodelski:
@tmcgilchrist I'd be happy to add some stuff, e.g. comments and naming. Could you provide me with write access to this branch?

src/lf_stack.ml
bartoszmodelski: DSCheck: 

```

 let test () = 
  let queue = create () in 
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  Atomic.spawn (fun () -> 
    clean_until queue (fun v -> v == 1));
  Atomic.spawn (fun () -> 
    clean_until queue (fun v -> v == 2));
  Atomic.final (fun () -> 
    Atomic.check (fun () -> 
      not (is_empty queue)))
```

src/lf_msqueue.ml
bartoszmodelski: DSCheck: 

```
 let test () = 
  let queue = create () in 
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  push queue 1; 
  push queue 2;
  Atomic.spawn (fun () -> 
    clean_until queue (fun v -> v == 1));
  Atomic.spawn (fun () -> 
    clean_until queue (fun v -> v == 2));
  Atomic.final (fun () -> 
    Atomic.check (fun () -> 
      not (is_empty queue)))
```

src/lf_stack.ml
bartoszmodelski: Ah no - impossible since we hold the reference to `s` the whole time. Marking as resolved.

bartoszmodelski:
I did some tests with backoffs. Somewhat unsurprisingly, backoff can have profound impact on performance. For example, it reduced runtime of my 12-thread CAS-contended benchmark by 90% (from 2s to <0.2s). But that gain depends heavily on tuning. Conversely, with the default constants, the improvement was around 10%, and that's not a crazy gain on such an artificial benchmark. 

We could try to improve the design of backoff module but it's kinda hard without any priors about the workload. So I'd vote to keep the backoff module and open up all the knobs to users (or even have structures take Backoff_intf).

bartoszmodelski:
Pushed benchmarks, relevant output below. 

```
    {
      "name": "backoff-on-basic",
      "metrics": [
        {
          "name": "time",
          "value": 0.214728,
          "units": "s",
          "description": "median time result"
        },
        {
          "name": "throughput",
          "value": 13971171.419245,
          "units": "item/s",
          "description": "median throughput result"
        }
      ]
    },
    { 
      "name": "backoff-off-basic",
      "metrics": [
        { 
          "name": "time",
          "value": 0.216608,
          "units": "s",
          "description": "median time result"
        },
        { 
          "name": "throughput",
          "value": 13849900.937789,
          "units": "item/s",
          "description": "median throughput result"
        }
      ]
    },
    { 
      "name": "backoff-on-artificial",
      "metrics": [
        { 
          "name": "time",
          "value": 0.143317,
          "units": "s",
          "description": "median time result"
        },
        { 
          "name": "throughput",
          "value": 20932620.214102,
          "units": "item/s",
          "description": "median throughput result"
        }
      ]
    },
    { 
      "name": "backoff-off-artificial",
      "metrics": [
        { 
          "name": "time",
          "value": 0.359622,
          "units": "s",
          "description": "median time result"
        },
        { 
          "name": "throughput",
          "value": 8342092.492263,
          "units": "item/s",
          "description": "median throughput result"
        }
      ]
    }

```

bartoszmodelski:
Treiber stack was mishandling the first node, so I corrected it to be closer to the wiki implementation. 

Replication: 
```
$ git checkout dc40de5  
$ dune utop
utop # let s = Lockfree.Triber_stack.create () in
Lockfree.Triber_stack.push s 123;
Lockfree.Triber_stack.pop s;;
- : int option = Some 0
```

bartoszmodelski:
Ok, this looks good to be merged to me. cc @tmcgilchrist 

* I've checked both structures with dscheck. It's especially useful for stack, since dscheck is the only place where we can validate lifo order. Tests attached (we can merge them after https://github.com/ocaml-multicore/lockfree/pull/34). 
* I'm holding off with benchmarks (other than for backoff module) as they will mostly reuse things added in https://github.com/ocaml-multicore/lockfree/pull/21 .

[test1.txt](https://github.com/ocaml-multicore/lockfree/files/10086808/test1.txt)
[test2.txt](https://github.com/ocaml-multicore/lockfree/files/10086809/test2.txt)

lyrm (commented):
Great work !

I have just found a small mistake that should not prevent the queue for working, but we should still correct it. Everything else is either about documentation or a really minor notes.

I am writing a few additional `qcheck` tests that I will push very soon.

src/michael_scott_queue.ml
```diff
@@ -0,0 +1,96 @@
...
+  in
+  loop ()
+
+let push q v =
+  let rec find_tail_and_enq curr_end node =
+    if Atomic.compare_and_set curr_end Nil node then ()
+    else
+      match Atomic.get curr_end with
+      | Nil -> find_tail_and_enq curr_end node
+      | Next (_, n) -> find_tail_and_enq n node
```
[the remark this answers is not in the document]
bartoszmodelski: I spend some time thinking about this, and if I understand it correctly, I think it would not work. If we return the curr_end and try to use it as value on line 64, then we only get one chance to do the tail update - any invocation succeeding afterwards will have a different tail value. So the following interleaving would leave break it: 
```
Assume two threads A, B

A: insert node (line 52)
B: insert node (line 52)
B: attempt cas (line 64) with curr_end value -> fail, since thr 1 still hasn't updated tail 
A: attempt cas (line 64) with curr_end value -> success

Now the tail value is one node behind, and any new thread won't be able to modify it, since they use curr_end as well.
```

The original algorithm acn do it because it also attempts tail-CAS on every iteration of `find_tail_and_enq` (line E13).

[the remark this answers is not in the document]
lyrm: You are right, Thanks !

PR #28, opened 2022-11-01
Fix bench json

bartoszmodelski:
Current output of `make bench` is not exactly what ocaml-bench expects. Fix that.


bartoszmodelski:
Fixed in https://github.com/ocaml-multicore/lockfree/pull/27

PR #32, opened 2022-11-03
Fix results not being in a json list

ElectreAAS:
This fixes the failure `Invalid_argument("Json: not a list")` on current-bench :)


ElectreAAS:
Didn't catch PR #28, this is redundant, sorry for the noise :/

PR #34, opened 2022-11-21
Dscheck meets lockfree

bartoszmodelski:
As in the title; integrate lockfree with dscheck, which is now in opam. 

Includes tests for:
* spsc queue
* mpsc queue
* ws deque


lyrm:
Thanks for the PR !

I don't think it is alright to have an indirection for the  `Atomic` library for anything else than testing. Maybe you could wrap the `Atomic` library as a virtual library with the default implementation being the `stdlib` one ?

lyrm:
Thanks ! Great work !

PR #36, opened 2022-11-24
Apply ocamlformat

bartoszmodelski:
Ocaml-ci is now on. Fix the build: 

* Add .ocamlformat
* Apply it


lyrm:
Thanks, I wanted to do that for a while ! 

Maybe use the lastest `.ocamlformat` version though (0.24.1) ?

bartoszmodelski:
`0.23.0` is the latest for 5.0.0+trunk - I think using newer might break compatibility with the coming OCaml 5 release

lyrm:
Good point.

PR #37, opened 2022-11-28
Enable OCaml 4 compatibility

Sudha247:
Uses [`domain_shims`](https://gitlab.com/gasche/domain-shims) to maintain compatibility with OCaml 4. A mock `Atomic` module is available in OCaml 4 from 4.12 onward. This is a prerequisite for a sequential version of Domainslib for OCaml 4.

I've reduced the count in spsc test, as the original count was taking > 20 minutes.

### DIFF
--- a/lockfree.opam
+++ b/lockfree.opam
@@ -11,6 +11,7 @@ tags: []
 depends: [
   "ocaml" {>= "5.0"}
   "dune" {>= "3.0"}
+  "domain_shims" {>= "0.1.0"}
   "qcheck" {with-test & >= "0.18.1"}
   "qcheck-alcotest" {with-test & >= "0.18.1"}
   "alcotest" {>= "1.6.0"}
--- a/lockfree.opam
+++ b/lockfree.opam
@@ -9,7 +9,7 @@ dev-repo: "git+https://github.com/ocaml-multicore/lockfree.git"
 bug-reports: "https://github.com/ocaml-multicore/lockfree/issues"
 tags: []
 depends: [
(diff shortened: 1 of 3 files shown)


test/test_spsc_queue.ml
```diff
@@ -22,14 +22,14 @@ let test_full () =
   print_string "test_spsc_queue_full: ok\n"
 
 let test_parallel () =
-  let count = 100_000 in
+  let count = 1000 in
```
bartoszmodelski, line 25: It would be nice to keep the higher number in general case. How about something like this: 

```
let count = 
  let ocaml_4 = Char.code (String.get Sys.ocaml_version 0) < Char.code '5' in
  if ocaml_4 then 1_000 else 100_000 
in 
```

Sudha247: Good idea. Done!

lyrm:
With the new addition of `dscheck` that requires OCaml 5 (and effects),  this PR does not work anymore as it is. @bartoszmodelski : do you have suggestions on  how to make it work?

lyrm:
@jmid suggested me a great idea to bypass the issue with `dscheck` with the  method used in [the repo](https://github.com/kit-ty-kate/ocaml-monomorphic ) that may make it possible to shadow the `Atomic` library without using a virtual library. I will be looking into it.

lyrm:
I merged with the last changes that remove dependency over `dscheck`.

@Sudha247 : could you check that everything is ok and merge if it is ?

Sudha247:
Looks good, thanks!

PR #38, opened 2022-11-28
Fix Ws_deque.push documentation

art-w:
The `Ws_deque` actually behaves like a stack for the owner of the deque!


lyrm:
Indeed. Thanks !

PR #40, opened 2022-12-01
Avoid TSan warning in the test-suite

fabbing:
In the continued joint effort with @OlivierNicole to understand the TSan crash in the testsuite, we propose a few changes:
- Avoid the risk of overflowing TSan shadow task (causing a SEGV) when too many '@' calls are nested by using tail-recursive calls.
- Avoid some data race detection in the tests that wait for the other domain by replacing the polling of `Semaphore.Counting.get_value` with a `Semaphore.Counting.try_acquire`.

### DIFF
--- a/test/qcheck_mpsc_queue.ml
+++ b/test/qcheck_mpsc_queue.ml
@@ -434,7 +434,8 @@ let tests_one_consumer_two_producers =
 
           let multi_push lpush =
             Semaphore.Counting.acquire sema;
-            while Semaphore.Counting.get_value sema <> 0 do
+            while Semaphore.Counting.try_acquire sema do
+              Semaphore.Counting.release sema;
               Domain.cpu_relax ()
             done;
             try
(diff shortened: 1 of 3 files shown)


bartoszmodelski:
Nice find! Implementation of semaphore is flawed, I'll shot a PR to fix it.

OlivierNicole:
Is it? It is specified in [the documentation](https://v2.ocaml.org/api/Semaphore.Counting.html) that the `get_value` function can cause races:

> get_value s returns the current value of semaphore s. The current value can be modified at any time by concurrent [Semaphore.Counting.release](https://v2.ocaml.org/api/Semaphore.Counting.html#VALrelease) and [Semaphore.Counting.acquire](https://v2.ocaml.org/api/Semaphore.Counting.html#VALacquire) operations. Hence, the get_value operation is racy, and its result should only be used for debugging or informational messages.

bartoszmodelski:
I see -- the changes look good to me

lyrm:
I did not realize `Semaphore.Counting.get_value` was racy. It is good to know ! 
Everything seems good to me too !

PR #43, opened 2022-12-12
Add STM ws_deque tests from multicoretests repo

jmid:
We now released `qcheck-stm` so here comes a PR with `STM` tests of `Ws_deque` from ocaml-multicore/multicoretests

STM tests are model-based and so describe the intended state of a `Ws_deque` by a list (`type state`).
A slight complication is the contract for `pop` and `push` which is only allowed by the deque owner.
We therefore use two generators - one for the generator and one for stealers.

From the model description we derive 3 tests:
- a sequential test of `push` and `pop`
- a parallel test with one owner and one stealer
- a negative parallel test that breaks the contract and hence fails as expected

### DIFF
--- a/lockfree.opam
+++ b/lockfree.opam
@@ -13,6 +13,7 @@ depends: [
   "dune" {>= "3.0"}
   "domain_shims" {>= "0.1.0"}
   "qcheck" {with-test & >= "0.18.1"}
+  "qcheck-stm" {with-test & >= "0.1"}
   "qcheck-alcotest" {with-test & >= "0.18.1"}
   "alcotest" {>= "1.6.0"}
   "yojson" {>= "2.0.2"}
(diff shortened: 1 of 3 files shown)


lyrm:
Thanks for the PR ! I have a lot of multicoretests for the lock-free data structures ready and waiting to be added !

jmid:
Hi @lyrm :wave:
Sorry, I didn't notice this was complaining about missing dependencies!
8f0ceb6 adds a hand-written `show`-function, so `ppx_deriving.show` is no longer a requirement - but I forgot to remove the ` (preprocess (pps ppx_deriving.show))` line from the `dune` rule :facepalm: 
(I don't believe the test from this PR needs `ppx_deriving_qcheck`)

lyrm:
Oh, thanks and sorry, I should  definitely have seen that !

PR #50, opened 2022-12-23
Mark benchmarks as optional library

bartoszmodelski:
This sets benchmarks as optional and puts them into a different new package (which includes yojson dependency).

With this change:
* installing `lockfree` (`opam install ./lockfree.opam --deps-only`) does not install yojson
* pipelines still work as they install all packages in the repo (something like `opam install . --deps-only` I suspect)

I decided to make the benchmarks package a superset of lockfree (so contributor can just install `lockfree-with-benchmarks` to get everything) but I'm happy to do it in a different way if there's a reason.

### DIFF
--- a/bench/dune
+++ b/bench/dune
@@ -1,3 +1,4 @@
 (executables
  (names main mpmc_queue_cmd)
- (libraries lockfree unix yojson))
+ (libraries lockfree unix yojson)
+ (optional))
--- a/lockfree-bench.opam
+++ b/lockfree-bench.opam
@@ -9,14 +9,12 @@ dev-repo: "git+https://github.com/ocaml-multicore/lockfree.git"
 bug-reports: "https://github.com/ocaml-multicore/lockfree/issues"
 tags: []
 depends: [
-  "ocaml" {>= "4.12"}
+  "ocaml" {>= "5.0"}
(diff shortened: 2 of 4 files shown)


bartoszmodelski:
[Example in Irmin](https://github.com/mirage/irmin/blob/main/irmin-bench.opam) - they seem to follow the alternative approach

polytypic:
I don't know what the recommended approach is with dune/opam, but I wonder whether making benchmarks tests (instead of ordinary executables) would be a suitable solution?  (And then the dependencies would be `with-test`.)

lyrm:
That would mean that benchmarks run with `opam runtest`. I am not sure it is a good idea. Other opinions are welcomed :)

polytypic:
Just noting that having benchmarks also function as tests is not necessarily bad.  That may help with making sure that both the library works and that the benchmarks also work.  Just need to make the default parameters such that running the benchmarks doesn't take a lot of time.

bartoszmodelski:
Sounds like it would work, but a bit less customary?

Sudha247:
`yojson` and `alcotest` have since been marked as test-only dependencies in #76. Thanks for the efforts to find what turned out to be a reasonable solution. Closing this PR.

Issue #44, opened 2022-12-16
Dscheck integration breaks utop 


bartoszmodelski:
Current integration of dscheck breaks utop ([link](https://discuss.ocaml.org/t/ocaml-5-0-0-alpha1-utop-require-domainslib/10969)), and there does not seem to be a simple way to reconcile the two (per [limitations](https://dune.readthedocs.io/en/stable/variants.html)). Is there a different way to achieve the same separation (i.e. no dscheck shadowing in production code) without losing utop? 
 
- A ppx that will select implementation of `Atomic.*` depending on environment flag 
- A different way to do the shadowing? (cc @lyrm)

Sudha247:
Can we make dscheck a `with-test` dependency? One need not install dscheck when installing lockfree then.

bartoszmodelski:
We chatted during the multicore meeting: 

1. Using `with-test` dependency sounds good to resolve the issue above. Nonetheless, it turns out the current integration also causes problems with OCaml 4 compatiblity (https://github.com/ocaml-multicore/lockfree/pull/37#issuecomment-1351369115), and there was some other idea proposed to try out there (https://github.com/ocaml-multicore/lockfree/pull/37#issuecomment-1351687468), so we might want to coordinate here to find one hack to kill both birds. 

2. Another, possibly the simplest solution, would be to just use a dune rule to copy dscheck's atomic in testing.

talex5:
Just to clarify, I have e.g.

- lib/mod.ml : uses `Atomic`
- test/dune : `(copy ../lib/mod.ml mod.ml)`
- test/atomic.ml  : `include Dscheck.TracedAtomic`

So mod.ml is compiled with the regular Atomic module in `lib`, and with `TracedAtomic` in `test`.

lyrm:
Solved by #53.

PR #53, opened 2023-01-09
Remove virtual library over atomic

lyrm:
In answer to issue #44.

I replaced the virtual library over `Atomic` with copy actions in `dscheck` dune rules as suggested by @talex5 in #44. I  also had to refactored the test directory as multiple data structures require the `Backoff` module and it was interfering with the fact that multiple dune rules can not use the same module. 

Also, it appears that `michael_scott_queue` and `treiber_stack` tests with `dscheck` are actually not running properly in the main branch as the virtual `Atomic` lib is not used. Once I correct that, it appears that some tests are overloading the memory most likely because there are too many paths to explore for `dscheck`.  I either simplified the tests or removed them (there are kind of redundant with previous ones).


talex5:
> I also had to refactored the test directory as multiple data structures require the Backoff module and it was interfering with the fact that multiple dune rules can not use the same module.

I had the same problem in Eio, but it turns out that using the plural forms (e.g. `executables` rather than several `executable` stanzas) solves it.

lyrm:
Oh, interesting ! I will remember that ! Thanks.

I think sorting the tests by data structures is pretty convenient as it enables to easily run tests for a single data structure but if anyone is against it, I will be happy to discuss it.

bartoszmodelski (approved):
Looks good to me. Perhaps I'd add a little comment in `atomic.ml` or markdown file in `test/` explaining that we copy-over implementations in order to shadow part of stdlib for dscheck tests.

lyrm:
Thanks for your review. That's a very good idea, so I just did it. Does it match what you had in mind ?

bartoszmodelski:
exactly that, thanks

Issue #47, opened 2022-12-21
Eio using MPSC queue?


kayceesrk:
https://github.com/ocaml-multicore/lockfree/blob/8c87c8bc675ac7537992e857705b7ee3a65ee2db/README.md?plain=1#L16

It is mentioned in the README.md that eio uses the MPSC queue. I didn't see `lockfree` being a dependency for `eio`: https://opam.ocaml.org/packages/eio/. Does that mean that `eio` has a copy of this data structure? If the API is compatible, can eio use the MPSC queue provided by `lockfree` @talex5?

Sudha247:
Yes, Eio should ideally use it from lockfree. We talked about this in the team meeting yesterday and @talex5 agreed. cc/ @lyrm

talex5:
> Does that mean that eio has a copy of this data structure?

lockfree copied it from Eio. And without adding my copyright to its license file (see https://github.com/ocaml-multicore/eio/blob/82886514b710f9fc4fec133e53db7f4772c66a88/LICENSE.md).

This implementation was specialised for Eio by dropping support for multiple consumers. But for the lockfree project it might be better to provide the more general version.

kayceesrk:
Thanks @talex5. Can we fix the LICENSE please @lyrm @Sudha247?

Sudha247:
We should fix the license bit.

Lockfree itself doesn't depend on `yojson`. It's needed for integrating the benchmarks with `ocaml-cb`. I wonder if we can restrict it to be an optional dependency (cc @bartoszmodelski).

bartoszmodelski:
Created https://github.com/ocaml-multicore/lockfree/pull/50

PR #54, opened 2023-01-23
License issue

lyrm:
To correct license issue raised in #47.


Sudha247:
cc @talex5 - do we need to add anything?

talex5:
Looks good to me - thanks!

PR #55, opened 2023-01-24
Update README

lyrm:
This PR updates documentation with :
- README : more contexts about lock freedom and this library
- README : a more recent usage example
- test/README : some notes about the various kind of tests that are done here
- CONTRIBUTING : some hints on what is expected of a PR for a new data structure.

Any comment, addition, English correction or improvement is obviously welcomed !

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,21 @@
+## Contributing
+Contributions of more lockfree data structures are appreciated! Please create issues/PRs to this repo.
+
+### Maintainers
+The current list of maintainers is as follows:
+
+-   @bartoszmodelski Bartosz Modelski
+-   @kayceesrk KC Sivaramakrishnan
+-   @lyrm Carine Morel
+-   @Sudha247 Sudha Parimala
+
+### Guidelines  for new data structures implementation
+Reviewing a lock free implementation takes times. Here are a few guidelines to make it easier for the reviewers :
+- choose a needed data structure (listed in issues)
+- implement a well know algorithm (there are a lot !)
+	- from a *reviewed* paper, ideally with proof of lock freedom
+	- from a well known and used concurrent library (like `java.util.concurrent`)
+-  write tests with **multiple** domains. All the following tests are expected to be provided before a proper review is done, especially for implementation that does not come from a well-know algorithm :
+	- unitary tests and `qcheck tests` : with one and multiple domains. If domains have specific roles (producer, consumer, stealer, etc..), it should appear in the tests.
+	- tests using `STM` from `multicoretest`
+	- (*optional*) `dscheck` tests
--- a/README.md
+++ b/README.md
@@ -1,40 +1,81 @@
 # `lockfree` — Lock-free data structures for Multicore OCaml
 --------------------------------------------------------
+## What is in this library ?
+This library is a collection of Concurrent Lockfree Data Structures for OCaml 5. It contains:
 
-A collection of Concurrent Lockfree Data Structures for OCaml 5. It contains:
+| Name | What is it ? | Sources |
+|-------|---------|----------|
+| [Treiber Stack](src/treiber_stack.mli) | A classic multi-producer multi-consumer stack, robust and flexible. Recommended starting point when needing LIFO structure | |
+| [Michael-Scott Queue](src/michael_scott_queue.mli) | A classic multi-producer multi-consumer queue, robust and flexible. Recommended starting point when needing FIFO structure. | [Simple, Fast, and Practical Non-Blocking and Blocking Concurrent Queue Algorithms](https://www.cs.rochester.edu/~scott/papers/1996_PODC_queues.pdf)|
+| [Chase-Lev Work-Stealing Deque](src/ws_deque.mli) | Single-producer, multi-consumer dynamic-size double-ended queue (deque). Ideal for throughput-focused scheduling using per-core work distribution. Note, `pop` and `steal` follow different ordering (respectively LIFO and FIFO) and have different linearization contraints. | [Dynamic circular work-stealing deque](https://dl.acm.org/doi/10.1145/1073970.1073974) and [Correct and efficient work-stealing for weak memory models](https://dl.acm.org/doi/abs/10.1145/2442516.2442524)) |
+| [SPSC Queue](src/spsc_queue.mli) | Simple single-producer single-consumer fixed-size queue. Thread-safe as long as at most one thread acts as producer and at most one as consumer at any single point in time. | |
+| [MPMC Relaxed Queue](src/mpmc_relaxed_queue.mli) | Multi-producer, multi-consumer, fixed-size relaxed queue. Optimised for high number of threads. Not strictly FIFO. Note, it exposes two interfaces: a lockfree and a non-lockfree (albeit more practical) one. See the mli for details. | |
+| [MPSC Queue](src/mpsc_queue.mli) | A multi-producer, single-consumer, thread-safe queue without support for cancellation. This makes a good data structure for a scheduler's run queue. It is used in [Eio](https://github.com/ocaml-multicore/eio). | It is a single consumer version of the queue described in [Implementing lock-free queues](https://people.cs.pitt.edu/~jacklange/teaching/cs2510-f12/papers/implementing_lock_free.pdf). |
 
-* [Treiber Stack](src/treiber_stack.mli) A classic multi-producer multi-consumer stack, robust and flexible. Recommended starting point when needing LIFO structure. 
-  
-* [Michael-Scott Queue](src/michael_scott_queue.mli) A classic multi-producer multi-consumer queue, robust and flexible. Recommended starting point when needing FIFO structure. It is based on [Simple, Fast, and Practical Non-Blocking and Blocking Concurrent Queue Algorithms](https://www.cs.rochester.edu/~scott/papers/1996_PODC_queues.pdf).
+## Lockfree ?
+### Definition
+[Wikipedia definition](https://en.wikipedia.org/wiki/Non-blocking_algorithm#Lock-freedom) is pretty clear. Briefly (and a bit inaccurately), it says that lock-freedom means:
+- obstruction freedom : any domain running in *isolation** for long enough can finish on its own
+- guarantees overall progress : at least one domain is doing progress and will finish its call in a finite number of steps.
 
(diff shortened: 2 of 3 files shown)


lyrm:
Some of the links are broken for now because the linked files are not pushed yet (there are `stm` tests). I am going to do the PR for these tests very soon.

polytypic:
Any chance we could start using some automated formatting tool on the `.md` files?  Not just in `lockfree`, but other `ocaml-multicore` projects.  I personally mostly use [prettier](https://prettier.io/), but anything that is easy to integrate to various editors (I mostly care about Emacs and VS Code, personally) would be fine with me.

README.md
```diff
@@ -1,40 +1,81 @@
...
+- guarantees overall progress : at least one domain is doing progress and will finish its call in a finite number of steps.
 
-* [Chase-Lev Work-Stealing Deque](src/ws_deque.mli) Single-producer, multi-consumer dynamic-size double-ended queue (deque) (see [Dynamic circular work-stealing deque](https://dl.acm.org/doi/10.1145/1073970.1073974) and [Correct and efficient work-stealing for weak memory models](https://dl.acm.org/doi/abs/10.1145/2442516.2442524)). Ideal for throughput-focused scheduling using per-core work distribution. Note, [pop] and [steal] follow different ordering (respectively LIFO and FIFO) and have different linearization contraints.
+**In isolation* : a domain is said to run *in isolation* when it is the only domain to take steps. The other domains do nothing during that time even if they are in the middle of a method call. It is used to described properties, but not an actual domain behaviour.
 
-* [SPSC Queue](src/spsc_queue.mli) Simple single-producer single-consumer fixed-size queue. Thread-safe as long as at most one thread acts as producer and at most one as consumer at any single point in time.
+### How does it works ?
```
Sudha247, line 23:
```suggestion
### How does it work?
```

README.md
```diff
@@ -1,40 +1,81 @@
...
+### Why use lock free data structures ?
+#### In brief
+| Pros |  Cons |
+|------|-------|
+| thread safe | **not** starvation free |
+| non blocking and dead-lock safe |  not easily composable (see [reagents](https://github.com/ocaml-multicore/reagents)) |
+| usually good performances as it runs in parallel most often than blocking data structures | still subject to contention issues |
 
-* [MPSC Queue](src/mpsc_queue.mli) A multi-producer, single-consumer, thread-safe queue without support for cancellation. This makes a good data structure for a scheduler's run queue. It is used in [Eio](https://github.com/ocaml-multicore/eio). It is a single consumer version of the queue described in [Implementing lock-free queues](https://people.cs.pitt.edu/~jacklange/teaching/cs2510-f12/papers/implementing_lock_free.pdf).
+#### What does that means ?
```
Sudha247, line 34:
```suggestion
#### What does that mean?
```

README.md
```diff
@@ -1,40 +1,81 @@
...
+If you need a *non-blocking* data structure, you are at the right place. However, if you only need a *thread safe* one you may not find what you need here. Especially, lock free data structures does not necessary have better performances than blocking ones. However,  it ensures that is a thread is delayed, that won't delay the other ones (this is the obstruction freedom property)
```
Sudha247, line 35:
```suggestion
If you need a *non-blocking* data structure, you are at the right place. However, if you only need a *thread safe* one you may not find what you need here. Especially, lock free data structures do not necessary have better performances than blocking ones. However, it ensures that if a thread is delayed, that won't delay the other ones (this is the obstruction freedom property).
```

README.md
```diff
@@ -1,40 +1,81 @@
...
 
-## Usage
+Also lock freedom is a restraining property, which means it is not easy (and it may not even be possible), to modify or add a function to a lock free data structure while maintaining this property. That explains why most of time the provided data structure are a very simple API.
```
Sudha247, line 37:
```suggestion
Also lock freedom is a restraining property, which means it is not easy (and it may not even be possible), to modify or add a function to a lock free data structure while maintaining this property. That explains why most of time the provided data structures have a very simple API.
```

test/README.md
```diff
@@ -1,5 +1,30 @@
-## About dscheck tests
+## Tests in `lockfree`
+Several kind of tests are provided for each data structure:
+- unitary tests and `qcheck` tests: check semantics and expected behaviors with one and more domains;
+- `STM` tests: check _linearizability_ for two domains (see [multicoretests library](https://github.com/ocaml-multicore/multicoretests));
+- `dscheck`: checks _non-blocking_ property for as many domains as wanted (for two domains most of the time). It is limited to the paths explored by the tests.
 
+### Unitary parallel tests and  `QCheck` tests
+Every data structures should have separate parallel tests for all core operations and for most useful combinations of them to make sure things work the way you expected them to, even in parallel.
```
Sudha247, line 8:
```suggestion
Every data structure should have separate parallel tests for all core operations and for most useful combinations of them to make sure things work the way you expected them to, even in parallel.
```

test/README.md
```diff
@@ -1,5 +1,30 @@
...
+
+Inside the parallel tests, it's important to ensure there's a high chance of actual parallel execution.  Keep in mind that spawning a domain is an expensive operation: if a domain only launches a few simple operations,  it will most likely run in isolation.  Here are a few tricks to ensure your tests actually runs with parallelism:
+-  make a domain repeat its operations . In particular, that usually works well when testing a single function.
+-  use a semaphore to make all domains wait for each other before starting (see [these tests for example) ](ws_deque/qcheck_ws_deque.ml)).
+- if you are still not sure your tests have a good chance to run in parallel, try it in the top level, and use print or outputs to understand what is happening.
+
+### Linearizability test with `STM`  (see [multicoretests](https://github.com/ocaml-multicore/multicoretests))
+
+### How to write `STM` tests ?
+`STM` tests works by comparing the results of two domains each executing a random list of method calls to a sequential model provided by the user. Most of the time, these tests are easy and quick to write and can catch a lot of bugs.
```
Sudha247, line 18:
```suggestion
`STM` tests work by comparing the results of two domains each executing a random list of method calls to a sequential model provided by the user. Most of the time, these tests are easy and quick to write and can catch a lot of bugs.
```

CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
...
+The current list of maintainers is as follows:
+
+-   @bartoszmodelski Bartosz Modelski
+-   @kayceesrk KC Sivaramakrishnan
+-   @lyrm Carine Morel
+-   @Sudha247 Sudha Parimala
+
+### Guidelines  for new data structures implementation
+Reviewing a lock free implementation takes times. Here are a few guidelines to make it easier for the reviewers :
+- choose a needed data structure (listed in issues)
```
Sudha247, line 14: I'd rephrase this as - "The issue tracker has a good list of data structures to choose from" - it could be the case that the contributor has uses of their own not captured in our issue tracker.

Sudha247 (commented):
Thanks for updating the `README` and `CONTRIBUTING` guide! I've left some comments below.

CONTRIBUTING.md
```diff
@@ -0,0 +1,21 @@
+## Contributing
+Contributions of more lockfree data structures are appreciated! Please create issues/PRs to this repo.
```
bartoszmodelski, line 2: Are we sure we want to restrict this repo to be lock-free only and have another, concurrent one? It does not seem like a natural boundary to me, and I struggle to see how that benefits community or us. Here's a bunch of reasons:
* Many structures are partly lock-free. Are we going to have them duplicated in both (since only concurrent repo can have the extra feature/performance)? 
* We're going to spend a lot of time arguing over lock-freedom. It is not trivial.
    * Any SPMC/SPSC/MPSC structure we have are either incorrect or not lock-free formally. 
    * We had a test run with MPMC queue. It took many days and multiple people voicing opinions to decide even though it came from a paper. That's because it used a way of showing lock-freedom, which is common in the literature, but does not flow from wikipedia definition. 
* Testing and benchmarking for lock-free and "just" concurrent is the same (e.g. obstruction-free still requires lock-free tooling like dscheck). We'll have to dupe them in both repos, and they makes the setups prone to drifting apart. This is bad, as we really need a robust way of comparing fine-grained and locked structures. 
* It's not typical and it's going to be extra confusing for users. 

Is there a good reason to accept this split? 

I imagine there's some target demographics for lock-freedom (or, in particular, wait-freedom) like real-time devs, but we do have a big non-lock-free gc that can kick in at any safe-point. And then, such a demographic can be served by making a table of provided property in the README (lockfree, lockfree if precondidtion, lockfree if at most one consumer, ..).

Sudha247:
> Are we sure we want to restrict this repo to be lock-free only and maintain another, all-concurrent one?

Thanks for raising this question, this is something I've been thinking about too. I'm not opposed to this repo hosting other concurrent data structures. Partially lock-free structures don't fall under the strict definition of lock-free, but end up being useful. Users might also find lock-based data structures sufficient for their use-cases, as long as it is correct and decently performant.

Perhaps we could rename this repo as `concurrent`, and have a sub-library for strictly lockfree structures as something like `concurrent-lockfree`. Alternatively, we could have one single library and strictly lock-free structures can be tagged as `lf`, though I imagine this would mean more changes to dependencies. Doing this now won't affect much of the reverse-deps, as there aren't that many to begin with - this situation may change in a few months and renaming might be more tedious then. Domainslib and Eio depend on lockfree, they're both not at 1.0 yet so breaking changes should be okay. If there are clear benefits to keeping lock-free and other concurrent structures separate, happy to go with it.

lyrm:
> Any chance we could start using some automated formatting tool on the `.md` files? Not just in `lockfree`, but also other `ocaml-multicore` projects. I personally mostly use [prettier](https://prettier.io/), but anything that is easy to integrate to various editors (I mostly care about Emacs and VS Code, personally) would be fine with me. Also, tools like [pre-commit](https://pre-commit.com/) can further help to keep things clean.

Thanks for the tips, I will make an update with `prettier`.

lyrm:
@Sudha247 @bartoszmodelski : Thanks for your thoughts about keeping the repo lockfree only.  I agree with all your good reasons to not keep it `lockfree` only.

I think changing the name to `concurrent` is the most appropriate path, and then we could,  as`Java.concurrent` does , have a proper documentation explaining what data structure has which progress properties (non blocking, lock-free, lock-based etc..).

lyrm:
Cancelling this, as a refactoring is in progress.

PR #60, opened 2023-03-05
Lock-free sets and computing size of a lock-free structure

nikochiko:
Computing the size of a concurrent structure is a non-trivial problem.
There has been a research paper recently on a lockfree, linearizable size operation on concurrent data structures [[Sela, Petrank. 2022](https://arxiv.org/pdf/2209.07100.pdf)].

This PR starts with the implementation of a lock-free set in OCaml, which will be a good first step, and will give an idea of how a size operation can be implementation for more complex data structures.

The set implementation depends on [[Maged. 2022](https://people.csail.mit.edu/xchen/parallel-computing/hash-table.pdf)].

Project mentor: @Sudha247


nikochiko:
CI is failing because the set is built on a linked list with a dummy node. The `create` function initialises the list with a node with a value of 0 (int, whereas the set should be polymorphic).
I didn't use a  `Dummy` constructor because that was making the code more complex.

For now I have verified that all the methods work in concurrent settings. Once size is implemented, I plan to rewrite it with Dummy constructor or something else that doesn't assume a type like here.

nikochiko:
@Sudha247 I have added an implementation of the sets with support for size.

The `insert` and `delete` functions take a thread_id argument as an `int`. The metadata is stored as a hashtable, with each thread having two counters: one for insertions and one for deletions.

This worked for me in some basic runs. This should work properly in a concurrent setting.

The set is implemented as a sorted linked list, and the contained type must support comparisons.

The approach is this:
- Every thread has an insertions and deletions counter in metadata. insertions / deletions can only increase. we do a CAS with like `cas(counter, old_value, old_value+1)` after each update. we ignore the result because another thread could've already done this.
- Each node has a `value`, `next`, and `update_info` attribute.
  - `value` is `None` only or the anchor dummy node
  - if `next` is `None`, we have reached end of list
  - `update_info` stores the metadata updation that needs to be made for this object
- when we insert or delete, we always perform a CAS on the node whose "next" must be changed. this allows us to catch when a concurrent thread changes the "next" references or "update_info".
- threads update metadata with `update_info` on another thread's behalf.
- insertions write metadata after their node is added
- deletions write metadata before the node is unlinked
- threads help other threads to delete a node

Some limitations of this are that:
- if we create too many threads (or thread IDs), we will have more collisions and slower metadata updates when inserting and deleting. these updates are atomic, and we attempt it from each thread that reaches a recently inserted node
- the running thread must know its id to perform an insert/delete operation

TODOs:
- at the moment, the update_info is not being reset back to None after the metadata is updated. it should be straightforward: this only needs to be done for insertions, and it can be done with a cas from the thread that was originally responsible for the insertion. this doesn't affect the correctness of the program.

polytypic:
Thanks for referencing the [Concurrent Size](https://arxiv.org/pdf/2209.07100.pdf) paper!

PR #99 contains a new algorithm for concurrent size inspired by the paper.  The new algorithm uses the same snapshotting technique as in the paper.  Instead of having guaranteed unique counters per thread, the new new algorithm updates the counters transactionally such that a single operation (incr/decr) is done only once.  This makes the counter updates lock-free rather than wait-free.  Contention on counter updates can be adjusted via an optional `n_way` argument and with a large enough `n_way` argument the counter update should become practically wait-free.  Computing the size is still (essentially) wait-free (it is the same algorithm).  PR #99 also uses the algorithm to provide a skiplist with size.

lyrm:
I'm closing this PR as #102 was merged and proposed a more elaborated version of the algorithm. 

Thanks again for your contribution and pointing to the concurrent size article. We ended up with a very interesting new feature thanks to you  !

PR #61, opened 2023-03-06
STM tests 

lyrm:
Add `Multicore.STM` tests for :
- `spsc_queue`
- `treiber_stack`
- `michael_scott_queue`
-  `mpsc_queue`

### DIFF
--- a/test/michael_scott_queue/stm_michael_scott_queue.ml
+++ b/test/michael_scott_queue/stm_michael_scott_queue.ml
@@ -1,0 +1,67 @@
+(** Sequential and Parallel model-based tests of michael_scott_queue *)
+
+open QCheck
+open STM
+module Ms_queue = Lockfree.Michael_scott_queue
+
+module MSQConf = struct
+  type cmd = Push of int | Pop | Is_empty
+
+  let show_cmd c =
+    match c with
+    | Push i -> "Push " ^ string_of_int i
+    | Pop -> "Pop"
+    | Is_empty -> "Is_empty"
+
+  type state = int list
+  type sut = int Ms_queue.t
+
+  let arb_cmd _s =
+    let int_gen = Gen.nat in
+    QCheck.make ~print:show_cmd
+      (Gen.oneof
+         [
+           Gen.map (fun i -> Push i) int_gen;
+           Gen.return Pop;
+           Gen.return Is_empty;
+         ])
+
+  let init_state = []
+  let init_sut () = Ms_queue.create ()
+  let cleanup _ = ()
+
+  let next_state c s =
+    match c with
+    | Push i -> i :: s
+    | Pop -> ( match List.rev s with [] -> s | _ :: s' -> List.rev s')
+    | Is_empty -> s
+
+  let precond _ _ = true
+
+  let run c d =
+    match c with
+    | Push i -> Res (unit, Ms_queue.push d i)
+    | Pop -> Res (option int, Ms_queue.pop d)
+    | Is_empty -> Res (bool, Ms_queue.is_empty d)
+
+  let postcond c (s : state) res =
+    match (c, res) with
+    | Push _, Res ((Unit, _), _) -> true
+    | Pop, Res ((Option Int, _), res) -> (
+        match List.rev s with [] -> res = None | j :: _ -> res = Some j)
+    | Is_empty, Res ((Bool, _), res) -> res = (s = [])
+    | _, _ -> false
+end
+
+module MSQ_seq = STM_sequential.Make (MSQConf)
+module MSQ_dom = STM_domain.Make (MSQConf)
+
+let () =
+  let count = 500 in
+  QCheck_base_runner.run_tests_main
+    [
+      MSQ_seq.agree_test ~count
+        ~name:"STM Lockfree.Michael_scott_queue test sequential";
+      MSQ_dom.agree_test_par ~count
+        ~name:"STM Lockfree.Michael_scott_queue test parallel";
+    ]
--- a/test/mpsc_queue/stm_mpsc_queue.ml
+++ b/test/mpsc_queue/stm_mpsc_queue.ml
@@ -1,0 +1,112 @@
+(** Sequential and Parallel model-based tests of mpsc_queue *)
+
+open QCheck
+open STM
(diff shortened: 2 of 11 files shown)


test/michael_scott_queue/stm_michael_scott_queue.ml
```diff
@@ -0,0 +1,68 @@
...
+    | Pop -> Res (option int, Ms_queue.pop d)
+    | Is_empty -> Res (bool, Ms_queue.is_empty d)
+
+  let postcond c (s : state) res =
+    match (c, res) with
+    | Push _, Res ((Unit, _), _) -> true
+    | Pop, Res ((Option Int, _), res) -> (
+        match List.rev s with [] -> res = None | j :: _ -> res = Some j)
+    | Is_empty, Res ((Bool, _), res) -> (
+        match s with [] -> res | _ -> not res)
```
jmid, line 53: This could also be written: `res = (s = [])` - which is a bit shorter

test/mpsc_queue/stm_mpsc_queue.ml
```diff
@@ -0,0 +1,147 @@
...
+  let init_sut () = Mpsc_queue.create ()
+  let cleanup _ = ()
+
+  let next_state c (is_closed, s) =
+    match c with
+    | Push i ->
+        (is_closed, if not is_closed then i :: List.rev s |> List.rev else s)
+    | Push_head i -> (is_closed, if not (is_closed && s = []) then i :: s else s)
+    | Is_empty -> (is_closed, s)
+    | Pop -> ( (is_closed, match s with [] -> s | _ :: s' -> s'))
```
jmid, line 53: double parens here

test/mpsc_queue/stm_mpsc_queue.ml
```diff
@@ -0,0 +1,147 @@
...
+        else res = Ok ()
+    | Pop, Res ((Result (Option Int, Exn), _), res) -> (
+        match s with
+        | [] ->
+            if is_closed then res = Error Mpsc_queue.Closed else res = Ok None
+        | x :: _ -> res = Ok (Some x))
+    | Is_empty, Res ((Result (Bool, Exn), _), res) ->
+        if is_closed && s = [] then res = Error Mpsc_queue.Closed
+        else if s = [] then res = Ok true
+        else res = Ok false
```
jmid, line 82: these two lines can we written as `else res = Ok (s = [])` I think

test/treiber_stack/stm_treiber_stack.ml
```diff
@@ -0,0 +1,68 @@
...
+    | Pop -> Res (option int, Treiber_stack.pop d)
+    | Is_empty -> Res (bool, Treiber_stack.is_empty d)
+
+  let postcond c (s : state) res =
+    match (c, res) with
+    | Push _, Res ((Unit, _), _) -> true
+    | Pop, Res ((Option Int, _), res) -> (
+        match s with [] -> res = None | j :: _ -> res = Some j)
+    | Is_empty, Res ((Bool, _), res) -> (
+        match s with [] -> res | _ -> not res)
```
jmid, line 53: Or `res = (s=[])`

jmid (commented):
Happy to see some STM tests! :smiley: 

I took a pass over this and it LGTM! 
I commented on a few minor nits.

test/mpsc_queue/stm_mpsc_queue.ml
lyrm: This is actually done by `ocamlformat`. Pretty weird. I raised an issue in ocamlfomat repo.

lyrm:
Thanks for the review ! Some `STM` tests (like `spsc_queue`'s ones) are pretty slow, any idea why ?

jmid:
> Thanks for the review ! Some `STM` tests (like `spsc_queue`'s ones) are pretty slow, any idea why ?

Yes, I can see that one is consistently slow in the parallel version across the 4 GitHub CI runs.
Nothing stands immediately out to me in the code though... :thinking: 
There will be quite some `List.rev`ing - but so do the other tests (and they aren't taking as long).

One suggestion could be to run a quick and dirty `sudo perf top` (if you are on Linux) to get a sense of where time is being spent. You can also do a focused `sudo perf top <pid>` and even `perf record` to store the observations for separate analysis...

jmid:
OK, I spent a bit of time digging into this. 
It turns out to be caused by a too optimistic `par_len` parameter, causing the exponential interleaving search to blow too much up. I fixed this in https://github.com/ocaml-multicore/multicoretests/pull/55 for STM - and apply in 4655f69 the same fix here: reducing the maximum list length of parallel commands from 15 to 12.

jmid:
If the data structures are OK with having tests run in 3 different domains (creator/seq.prefix, producer, consumer) using the exported `WSDT_dom.agree_prop_par` property could save quite a few lines by omitting a locally defined `agree_prop_par`.

jmid:
Come to think of it, we should add something like `agree_prop_par_asym` to `qcheck-STM` as that would reduce much of this customization to just one-liners... :smile:

jmid:
The PR ocaml-multicore/multicoretests#315 to add `agree_prop_par_asym` has been merged, so I've now taken a pass over this PR to reduce boilerplate test code, which has resulted in a nice reduction :smiley: 

I've included
- a similar reduction of the existing `ws_deque` test
- a pin of the development version of `qcheck-stm` until the next opam release

lyrm:
Mmmh, I removed the `pin` in the `.opam` file on `qcheck-stm` because I thought it was working without it, but it does not.

@jmid : any idea when a new version of `multicoretests` will be released with theses changes ?

jmid:
> Mmmh, I removed the `pin` in the `.opam` file on `qcheck-stm` because I thought it was working without it, but it does not.

Indeed. The added bindings are not available in the previous release, hence the pin.
This is a one-liner change once I roll release, so I don't see why this should hold back the PR.

> @jmid : any idea when a new version of `multicoretests` will be released with theses changes ?

Perhaps next week. I'm currently trying to push out a new release of the underlying QCheck first, but opam-ci is on its knees. If my luck continues like this, you may have to wait longer.

PR #62, opened 2023-03-07
Lock free hash table

lyrm:
### Last changes (16/05/23) : 
-  partially apply reviews (thanks @bartoszmodelski)
- add part of the documentation
- some renaming (still some more to do, as suggested in the review)
- wait-free `mem` and `find_opt`

### Still to do: 
- [ ] documentation
- [ ] add README note
- [ ] benchmarks
- [ ] finish cleaning the resizable htbl

In the implementation itself :
- [ ] add hash function
- [x] ~~wait-free mem function (may be keep both : a wait-free one and the current one that is helping with removal as they may be both useful)~~ : DONE 
- [x] ~~(optional) removal loop in the Llist.find function as written in the LockfreeList class in the AoMP book~~ : not improving performances.
 - [ ] (optional) changing the criteria to determine when to resize to reduce contention (currently every add and remove change a single value).

### For another PR 
Currently this hash table is actually extensible and not really resizable. It can only grow and has a maximum size. There are ideas on the book about how to make it resizable without limit (not sure about shrinking though).

### DIFF
--- a/bench/hmap.ml
+++ b/bench/hmap.ml
@@ -1,0 +1,147 @@
+module type Interface = sig
+  type 'a t
+  type key = int
+
+  val create : size_exponent:int -> 'a t
+  val add : key -> 'a -> 'a t -> bool
+  val find_opt : key -> 'a t -> 'a option
+  val remove : key -> 'a t -> bool
+end
+
+module Locked : Interface = struct
+  type key = int
+  type 'a t = (key, 'a) Hashtbl.t
+
+  let mutex = Mutex.create ()
+
+  let with_mutex f =
+    Mutex.lock mutex;
+    let v = f () in
+    Mutex.unlock mutex;
+    v
+
+  let create ~size_exponent =
+    let size = Int.shift_left 1 size_exponent in
+    Hashtbl.create size
+
+  let add (k : key) (v : 'a) (t : 'a t) =
+    with_mutex (fun () ->
+        Hashtbl.replace t k v;
+        true)
+
+  let find_opt k t = with_mutex (fun () -> Hashtbl.find_opt t k)
+
+  let remove k t =
+    with_mutex (fun () ->
+        Hashtbl.remove t k;
+        true)
+end
+
+let implementations =
+  [
+    ("locked\t\t\t", (module Locked : Interface));
+    ("lockfree\t\t", (module Lockfree.Hshtbl : Interface));
+    ("lockfree-resizable\t", (module Lockfree.Hshtbl_resizable : Interface));
+  ]
+
+(* module Hashtable = HashtblLocked *)
+module Hashtable = Lockfree.Hshtbl
+
+let run_bench item_count total_domains size_exponent preload_item_count workload
+    (impl_name, implementation) =
+  let module Hashtable = (val implementation : Interface) in
+  Random.init 0;
+
+  let hashtbl = Hashtable.create ~size_exponent in
+  let orchestrator = Orchestrator.init ~total_domains ~rounds:10 in
+
+  let key_range =
+    let size = Int.shift_left 1 size_exponent in
+    max size preload_item_count
+  in
+  for _ = 1 to preload_item_count do
+    Hashtable.add (Random.int key_range) 0 hashtbl |> ignore
+  done;
+
+  (* start domains *)
+  let domains =
+    List.init total_domains (fun _ ->
+        Domain.spawn (fun () ->
+            Orchestrator.worker orchestrator (fun () ->
+                for _ = 1 to item_count do
+                  let key = Random.int key_range in
+                  match workload with
+                  | `Add_remove ->
+                      Hashtable.add key 0 hashtbl |> ignore;
+                      Hashtable.remove key hashtbl |> ignore
+                  | `Get -> Hashtable.find_opt key hashtbl |> ignore
+                done)))
+  in
+  (* run test *)
+  let times = Orchestrator.run orchestrator in
+  List.iter Domain.join domains;
+  let time_median = Stats.median times in
+  let time_stddev = Stats.stddev times in
+  Printf.printf "[%s] time median: %f, stddev: %f\n" impl_name time_median
+    time_stddev
+
+let run item_count domains size_exponent preload_item_count workload =
+  List.iter
+    (run_bench item_count domains size_exponent preload_item_count workload)
+    implementations
+
+open Cmdliner
+
+let preload_item_count =
+  let default = 500 in
+  let info =
+    Arg.info [ "p"; "preload-items" ] ~docv:"INT"
+      ~doc:"Number of items to place in the hashmap before starting benchmark."
+  in
+  Arg.value (Arg.opt Arg.int default info)
+
+let size_exponent =
+  let default = 10 in
+  let info =
+    Arg.info
+      [ "s"; "init-size-exponent" ]
+      ~docv:"INT" ~doc:"Initial size (size = 2^parameter)."
+  in
+  Arg.value (Arg.opt Arg.int default info)
+
+let item_count =
+  let default = 100_000 in
+  let info =
+    Arg.info [ "i"; "item-count" ] ~docv:"INT"
+      ~doc:"Number of items to add and remove."
+  in
+  Arg.value (Arg.opt Arg.int default info)
+
+let domains =
+  let default = 4 in
+  let info =
+    Arg.info [ "d"; "domains" ] ~docv:"INT" ~doc:"Number of domains."
+  in
+  Arg.value (Arg.opt Arg.int default info)
+
+let workload =
+  let default = `Add_remove in
+  let info =
+    Arg.info [ "w"; "workload" ] ~docv:"[add-remove|get]"
+      ~doc:
+        "Workload to bench. The Get option might benefit from combination with \
+         preloading items to create varying hit rates."
+  in
+  Arg.value
+    (Arg.opt
+       (Arg.enum [ ("add-remove", `Add_remove); ("get", `Get) ])
+       default info)
+
+let cmd =
+  let open Term in
+  const run $ item_count $ domains $ size_exponent $ preload_item_count
+  $ workload
+
+let () =
+  exit @@ Cmd.eval
+  @@ Cmd.v (Cmd.info ~doc:"Hashmap Benchmark" "hmap_benchmark") cmd
--- a/src/htbl.ml
+++ b/src/htbl.ml
@@ -1,0 +1,354 @@
+(*
+ * Copyright (c) 2023, Carine Morel <carine.morel.pro@gmail.com>
+ *)
+
+module Type = struct
+  type 'b kind = Dummy | Regular of 'b
+  type key = int
+end
+
+module Llist = struct
+  include Type
+
+  type 'a marked = Last | LRemove | Normal of 'a node | Remove of 'a node
+  and 'a node = { key : key; value : 'a kind; next : 'a marked Atomic.t }
+
+  type 'a t = 'a marked Atomic.t
+
+  type 'a local = {
+    prev : 'a marked Atomic.t;
+    curr : 'a marked;
+    next : 'a marked;
+  }
+
+  let init () : 'a t = Atomic.make Last
+
+  let convert_to_normal = function
+    | (Last | Normal _) as node -> node
+    | LRemove -> Last
+    | Remove node -> Normal node
+
+  let mark_to_be_removed = function
+    | Last -> LRemove
+    | Normal node -> Remove node
+    | _ as x -> x
+
+  let rec find_loop key t prev curr =
+    match curr with
+    | Last | LRemove -> (false, { prev; curr; next = Last })
+    | Normal node | Remove node -> (
+        if Atomic.get prev != curr then try_again key t
+        else
+          let next = Atomic.get node.next in
+          match next with
+          | Normal _ | Last ->
+              if node.key >= key then (node.key = key, { prev; curr; next })
+              else find_loop key t node.next next
+          | _ ->
+              let next = convert_to_normal next in
+              if Atomic.compare_and_set prev curr next then
+                find_loop key t prev next
+              else try_again key t)
+
+  and try_again key t = find_loop key t t (Atomic.get t)
+
+  let find key t : bool * 'a local = try_again key t
+
+  let rec unsafe_add (key : key) (value : 'a kind) t : bool * 'a local =
+    let is_found, local = find key t in
+    if is_found then (false, local)
+    else if
+      Atomic.compare_and_set local.prev local.curr
+      @@ Normal { key; next = Atomic.make local.curr; value }
+    then (true, local)
+    else unsafe_add key value t
+
+  let add (key : key) (value : 'a kind) (t : 'a t) =
+    fst (unsafe_add key value t)
+
+  let rec unsafe_remove (key : key) t =
+    let is_found, local = find key t in
+    if not is_found then (false, local)
+    else
+      let curr =
+        match local.curr with
+        | Normal node -> node
+        | _ -> failwith " Should never happen"
+      in
+      if
+        not
+          (Atomic.compare_and_set curr.next local.next
+             (mark_to_be_removed local.next))
+      then unsafe_remove key t
+      else if Atomic.compare_and_set local.prev local.curr local.next then
+        (true, local)
+      else (
+        ignore (find key t);
+        (true, local))
+
+  let remove (key : key) (t : 'a t) = fst (unsafe_remove key t)
+  let mem key t = fst @@ find key t
+end
+
+module Common = struct
+  (* Key value used in the linked list are computed with [reverse] for the
+        dummy node and with [compute_hkey] for the regular node.
+
+     Example of a linked list, simplified to 8 bits integer. Only key
+     are written, as values of regular nodes do not change the linked
+     list order.
+
+      [ 0000 0000 (0, dummy) ]
+     -> [ 0001 0001 (8, regular)]
+     -> [ 0100 0000 (2, dummy)]
+     -> [ 0101 0001 (10, regular)]
+     -> [ 0110 1001 (22, regular)]
+     -> [ 1000 0000 (1, dummy)]
+     -> [ 1000 1001 (17, regular)]
+     -> [ 1100 0000 (3, dummy)]
+  *)
+  let reverse x =
+    (* works for int32 *)
+    let x = x land 0xff_ff_ff_ff in
+    let x = ((x land 0xaa_aa_aa_aa) lsr 1) lor ((x land 0x55_55_55_55) lsl 1) in
+    let x = ((x land 0xcc_cc_cc_cc) lsr 2) lor ((x land 0x33_33_33_33) lsl 2) in
+    let x = ((x land 0xf0_f0_f0_f0) lsr 4) lor ((x land 0x0f_0f_0f_0f) lsl 4) in
+    let x = ((x land 0xff_00_ff_00) lsr 8) lor ((x land 0x00_ff_00_ff) lsl 8) in
+    (x lsr 16) lor (x lsl 16) land 0xffffffff
+
+  let compute_hkey k = reverse k lor 0x00_00_00_01
+
+  (** unset most significant turn on bit (for int32) *)
+  let unset_msb key =
+    let a = key lor (key lsr 1) in
+    let a = a lor (a lsr 2) in
+    let a = a lor (a lsr 4) in
+    let a = a lor (a lsr 8) in
+    let a = a lor (a lsr 16) in
+    (a lsr 1) land key
+
+  let new_dummy_node id llist =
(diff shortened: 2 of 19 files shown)


src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+          let next = Atomic.get node.next in
+          match next with
+          | Normal _ | Last ->
+              if node.key >= key then (node.key = key, { prev; curr; next })
+              else find_loop key t node.next next
+          | _ ->
+              let next = convert_to_normal next in
+              if Atomic.compare_and_set prev curr next then
+                find_loop key t prev next
+              else try_again key t)
```
bartoszmodelski, line 51: If we fail here, then another thread must have succeeded at doing the deletion and, seemingly, we can just continue. Do we need to retry here?

src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+    let size = Int.shift_left 1 size_exponent in
+    let mask = size - 1 in
+    let llist = Llist.init () in
+    {
+      mask;
+      buckets =
+        Array.init size (fun i ->
+            if i = 0 then Atomic.make (new_dummy_node 0 llist)
+            else Atomic.make Llist.Last);
+    }
```
bartoszmodelski, line 158: Perhaps we can initialize the buckets on creation in the fixed case? It should improve the performance of hot-path further.

src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+
+  let find key { buckets; mask; _ } =
+    let is_found, local =
+      Llist.find (compute_hkey key) (get_bucket_ind key buckets mask)
+    in
+    if not is_found then None
+    else
+      match local.curr with
+      | Normal { value = Regular k; _ } | Remove { value = Regular k; _ } ->
+          Some k
```
bartoszmodelski, line 198: In the `Remove {..}` case the node is logically removed and is technically not in the linked list.

bartoszmodelski:
Looks pretty good! It needs a bit more work to be ready, but I expect the general design to hold up well in many scenarios. In particular, I ran a few tests with the fixed-array implementation against `Stdlib.Hashmap` (surrounded by a lock) and the lock-free hashtable tended to be multiple times faster with just 2 threads. 

I left a few comments in the code (focusing on linked list and non-resizable version for now). Also a couple more general ones:
* I'm wondering about naming here. If we want to call it hashtable, then it should probably be abbreviated to Hashtbl (to match `Stdlib.Hashtbl`). On the other hand, OCaml's hashtable has some somewhat surprising semantics (i.e. add hides existing binding) and we might want to differentiate. Perhaps just concurrent map would be better?
* Linked list (and the sorted insertion) is a useful structure. Can we expose it as a separate module? For example, sorted insertion + pop_front pave way to a simple implementation of priority queue.

polytypic:
> * I'm wondering about naming here. If we want to call it hashtable, then it should probably be abbreviated to Hashtbl (to match `Stdlib.Hashtbl`). On the other hand, OCaml's hashtable has some somewhat surprising semantics (i.e. add hides existing binding) and we might want to differentiate. Perhaps just concurrent map would be better?

I'd recommend not using the name `Hashtbl` unless the interface and semantics are meant to copy the `Stdlib.Hashtbl`.  In [kcas_data](https://ocaml-multicore.github.io/kcas/doc/kcas_data/Kcas_data/index.html#stdlib-style-data-structures), I used the names of the Stdlib structures (`Hashtbl`, `Queue`, and `Stack`) as they try to mimic the Stdlib structures to the point of being (more or less) drop-in replacements.  However, the Stdlib structures are not designed for concurrent programming.  They provide operations and have semantics that are difficult to implement efficiently in a lock-free manner.  So, I'm hoping that we might, at some point, have other related structures that are better suited for concurrent programming and lock-free implementation and at that point kcas_data could also provide composable versions of those.

polytypic:
>  I ran a few tests with the fixed-array implementation against Stdlib.Hashmap (surrounded by a lock) and the lock-free hashtable tended to be multiple times faster with just 2 threads.

Cool.  Which hash function did you use, BTW?

When I was working on the kcas_data `Hashtbl` I noticed that [the Stdlib generic `hash`](https://v2.ocaml.org/api/Hashtbl.html#1_Thepolymorphichashfunctions) (which is used by the `Stdlib.Hashtbl`) is relatively expensive.  IOW, skipping the hash function call when using integers as keys improved performance significantly. I'm intentionally [skipping the generic hash in my benchmark](https://github.com/ocaml-multicore/kcas/blob/main/test/kcas_data/hashtbl_bench.ml#L6), because I want to better isolate the costs of the `Hashtbl` implementation itself.

src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+    | Remove node -> Normal node
+
+  let mark_to_be_removed = function
+    | Last -> LRemove
+    | Normal node -> Remove node
+    | _ as x -> x
+
+  let rec find_loop key t prev curr =
+    match curr with
+    | Last | LRemove -> (false, { prev; curr; next = Last })
```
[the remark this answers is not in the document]
lyrm: Good question.

First important note : the `remove` information is for the node that precedes and not the node actually wearing it.

Then it is a question of use. For the linked list, we should never have a `LRemove` or `Remove node` in this pattern matching because the only case that could happen is at the first iteration. In any other cases, we would see it when inspecting the previous node, remove the previous node and convert the `Remove` node to a `Normal` one.  But there is no case where the first call of `find` on a a type `Llist.t` could encounter a `remove` node. So for the linked list that is an impossible case. 

For the hashtable however, the buckets point to dummy nodes in the middle of a linked list. So a dummy node can actually be marked if the node right before needs to be removed. As the `find` function begins its search at the marked dummy node, it has never encountered the node that actually needs to be removed and so can neither remove the previous node or unmarked the current one.

src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+    curr : 'a marked;
+    next : 'a marked;
+  }
+
+  let init () : 'a t = Atomic.make Last
+
+  let convert_to_normal = function
+    | (Last | Normal _) as node -> node
+    | LRemove -> Last
+    | Remove node -> Normal node
```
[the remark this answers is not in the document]
lyrm: A node is logically removed if its `next` field is wearing a `remove` mark.

````ocaml
let is_logically_removed node =  
  match Atomic.get node.next with 
  | LRemove  | Remove _ -> true
  |  _ -> false
````

File: src/htbl.ml [hunk 6]
```diff
@@ -0,0 +1,354 @@
...
+  let mark_to_be_removed = function
+    | Last -> LRemove
+    | Normal node -> Remove node
+    | _ as x -> x
+
+  let rec find_loop key t prev curr =
+    match curr with
+    | Last | LRemove -> (false, { prev; curr; next = Last })
+    | Normal node | Remove node -> (
+        if Atomic.get prev != curr then try_again key t
```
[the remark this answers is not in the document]
lyrm: That is making me think of a case that looks like it should not be working. I try it on `dscheck` that did not catch an issue. I will develop it here if I can resolve it myself.

src/htbl.ml
bartoszmodelski: (The hashtable is section 13.3 from The Art of Multiprocessor Programming.)

bartoszmodelski:
I've omitted the hash function as well and just used random ints (in fact, at the time I thought that `compute_hkey` is doing some built-in hashing).

src/htbl.ml
```diff
@@ -0,0 +1,354 @@
...
+      else grow t count
+
+  (** [add key value t] *)
+  let add (key : key) (value : 'a) (t : 'a t) =
+    let is_added =
+      Llist.add (compute_hkey key) (Regular value) @@ get_bucket key t
+    in
+    if not is_added then false
+    else (
+      Atomic.fetch_and_add t.count 1 |> grow t;
```
bartoszmodelski, line 323: For the record, I did some experimenting with this FAD. It looks like a potential bottleneck since all threads doing add/remove, which are mostly disjoint accesses, get serialized here. But it does not look to matter on 4 domain benchmarks (yet). It might scale differently on arm, where fad gets turned into load/store loop.

bartoszmodelski:
In the future we might want to think about adding a way to let user do safe updates. With just get/set/remove doing updates based on value is racy if keys are not unique. 

For example, suppose we have a map from PID to process owner. We want to delete all processes owned by X, so we inspect get the key and if owner = X, remove relevant PID from hashmap. But in the meantime another thread might have already deleted this entry because the process crashed and the PID has been reused for a different process. At the level of hashtable, we know that these two will be in different nodes.

src/htbl.ml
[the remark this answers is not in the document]
bartoszmodelski: FWIW I looked at the search method in pragmatic non-blocking linked-list paper. I'm not 100% comfortable with its details yet, but it is doing a similar kind of check.

src/htbl.ml
bartoszmodelski: Intuitively, I'm not loving this because its likely to generate extra contention when the structure is already busy. But as above, it's also done in the pragmatic linked list paper.

src/htbl.ml
lyrm: The paper I worked with is actually _Split-ordered lists: Lock-free extensible hash tables_ from O. Shalev and N. Shavit. I realized only a few weeks ago that it was better described and improved in _The Art of Multiprocessor Programming_. 

I will add the simplest algorithm improvements in this PR and will add the other ones in a later PR.  

The ones that I will include soon are : 
- (sure) wait-free `mem` function (may be keep both : a wait-free one and the current one that is helping with removal as they may be both useful)  
- (probably) removal loop in the `Llist.find` function as written in the `LockfreeList` class in the AoMP book
- (probably) changing the criteria to determine when to grow to reduce contention (currently every add and remove change a single value).

In another PR ; currently this hash table is actually extensible and not really resizable. It can only grow and has a maximum size. There are ideas on the book about how to make it resizable without limit (not sure about shrinking though).

polytypic:
I took the library of adding checkboxes to the PR description.  I kind of find them convenient and I guess it is not a super widely known feature.  Feel free to remove those.

c-cube:
Pardon my naive question, but is this going to stay specialized on integer keys? A functor to get arbitrary key types would make this a lot more general.

bartoszmodelski:
# Changes 
I've pushed two commits: 
* Added a small README note. 
* Cleaned up my benchmarks and pushed a unified cmd utility. It includes: 
   * Two workloads: add and remove item, check membership (find_opt)
   * Option to preload the structure with items  
   * All the other usual knobs 

# Benchmarks 
## 4 domains, add and remove random item 

```
dune build && ./_build/default/bench/hmap.exe --item-count 100000 --domains 4 --workload add-remove --preload 0 
[locked                 ] time median: 0.089562, stddev: 0.000312
[lockfree               ] time median: 0.010611, stddev: 0.000132
[lockfree-resizable     ] time median: 0.020596, stddev: 0.000123
```

As expected, we're a lot faster thank to true parallelism. 

## 1 domain, add and remove random item 
```
dune build && ./_build/default/bench/hmap.exe --item-count 100000 --domains 1 --workload add-remove --preload 0
[locked                 ] time median: 0.007261, stddev: 0.000148
[lockfree               ] time median: 0.005638, stddev: 0.000084
[lockfree-resizable     ] time median: 0.011077, stddev: 0.000082
```

This is slightly surprising to me. With just one domain the locked case should be super fast because I'd expect us to hit some fast-path in the mutex (1 atomic op?) and then use a much faster non-synched hashtable. As far as I remember we use pthread's mutex, so it might be that the cost of calling into C is quite high. 


## 2 domains, get with preload 1000 items (with 1024 being the size of hashtable)

```
dune build && ./_build/default/bench/hmap.exe --item-count 100000 --domains 2 --workload get --preload 1000
[locked                 ] time median: 0.018941, stddev: 0.000177
[lockfree               ] time median: 0.002990, stddev: 0.000052
[lockfree-resizable     ] time median: 0.004553, stddev: 0.000048
```
As expected, just doing `mem` is a lot faster in the lock-free cases.

## 2 domains, get with preload 100_000 (with 1024 being the size of hashtable) 

```
dune build && ./_build/default/bench/hmap.exe --item-count 100000 --domains 2 --workload get --preload 100000
[locked                 ] time median: 0.029811, stddev: 0.000197
[lockfree               ] time median: 0.098113, stddev: 0.000228
[lockfree-resizable     ] time median: 0.014080, stddev: 0.000130
```

Here, expectedly, the non-resizable version is performing poorly because each slot in the array ends up having a long linked list and that send lookup on long traversals. The resizable version is 2x faster than locked version. I'd expect to see more than 2x here, given that lock appears to be expensive in other benchmarks; we definitely scale a bit worse than the locked version.

bench/hmap.ml
```diff
@@ -0,0 +1,147 @@
...
+
+  let with_mutex f =
+    Mutex.lock mutex;
+    let v = f () in
+    Mutex.unlock mutex;
+    v
+
+  let create ~size_exponent =
+    let size = Int.shift_left 1 size_exponent in
+    Hashtbl.create size
```
polytypic, line 25: Note that this will use the generic hash which is very slow.

c-cube: agreed, if  you compare with `Hashtbl` you absolutely have to use `Hashtbl.Make`, for example with this:

```
type t = int
let equal : t -> t -> bool = (=)  (* type annotation might be important to avoid polymorphic compare *)
let hash = <your hash here>
```

polytypic: I would recommend using `Fun.id` as the hash function as I mentioned in a previous comment/question. The goal being to measure performance of the hash table intenals rather than the hash function.

polytypic:
Out of curiosity I tested how the [`Hashtbl`](https://ocaml-multicore.github.io/kcas/doc/kcas_data/Kcas_data/Hashtbl/index.html) implementation in [**kcas_data**](https://github.com/ocaml-multicore/kcas/#readme) compares.

Here is the diff:

```diff
modified   bench/dune
@@ -1,4 +1,4 @@
 (executables
  (names main mpmc_queue_cmd hmap)
- (libraries lockfree unix yojson cmdliner))
+ (libraries lockfree unix yojson cmdliner kcas_data))
 
modified   bench/hmap.ml
@@ -8,9 +8,17 @@ module type Interface = sig
   val remove : key -> 'a t -> bool
 end
 
+module Int = struct
+  include Int
+
+  let hash = Fun.id
+end
+
 module Locked : Interface = struct
+  module Hashtbl = Hashtbl.Make (Int)
+
   type key = int
-  type 'a t = (key, 'a) Hashtbl.t
+  type 'a t = 'a Hashtbl.t
 
   let mutex = Mutex.create ()
 
@@ -37,11 +45,33 @@ module Locked : Interface = struct
         true)
 end
 
+module Kcas_hashtbl = struct
+  open Kcas_data
+
+  type key = int
+  type 'a t = (key, 'a) Hashtbl.t
+
+  let create ~size_exponent =
+    let min_buckets = Int.shift_left 1 size_exponent in
+    Hashtbl.create ~min_buckets ~hashed_type:(module Int) ()
+
+  let add (k : key) (v : 'a) (t : 'a t) =
+    Hashtbl.replace t k v;
+    true
+
+  let find_opt k t = Hashtbl.find_opt t k
+
+  let remove k t =
+    Hashtbl.remove t k;
+    true
+end
+
 let implementations =
   [
     ("locked\t\t\t", (module Locked : Interface));
     ("lockfree\t\t", (module Lockfree.Hshtbl : Interface));
     ("lockfree-resizable\t", (module Lockfree.Hshtbl_resizable : Interface));
+    ("kcas_data\t\t", (module Kcas_hashtbl : Interface));
   ]
 
 (* module Hashtable = HashtblLocked *)
```

Here are the bench results:

```
+ dune exec --release -- ./bench/hmap.exe --item-count 100000 --domains 4 --workload add-remove --preload 0
[locked                 ] time median: 0.089719, stddev: 0.002471
[lockfree		] time median: 0.018099, stddev: 0.002830
[lockfree-resizable	] time median: 0.055691, stddev: 0.001710
[kcas_data		] time median: 0.049859, stddev: 0.005698
+ dune exec --release -- ./bench/hmap.exe --item-count 100000 --domains 1 --workload add-remove --preload 0
[locked                 ] time median: 0.006279, stddev: 0.000091
[lockfree		] time median: 0.006054, stddev: 0.000024
[lockfree-resizable	] time median: 0.008014, stddev: 0.000037
[kcas_data		] time median: 0.020660, stddev: 0.000068
+ dune exec --release -- ./bench/hmap.exe --item-count 100000 --domains 2 --workload get --preload 1000
[locked                 ] time median: 0.015168, stddev: 0.000361
[lockfree		] time median: 0.003168, stddev: 0.000042
[lockfree-resizable	] time median: 0.003699, stddev: 0.000014
[kcas_data		] time median: 0.002477, stddev: 0.000043
+ dune exec --release -- ./bench/hmap.exe --item-count 100000 --domains 2 --workload get --preload 100000
[locked                 ] time median: 0.023329, stddev: 0.000377
[lockfree		] time median: 0.059774, stddev: 0.000096
[lockfree-resizable	] time median: 0.008269, stddev: 0.000072
[kcas_data		] time median: 0.005082, stddev: 0.000079
```

Note that **kcas_data** `Hashtbl` is designed to be an almost drop in replacement for the Stdlib `Hashtbl`, so it definitely makes a few compromises (e.g. maintains length transactionally) when it comes to performance.  The **kcas_data** `Hashtbl` resizes automatically.  The implementation is fairly carefully made, but there are still at least a few potential optimizations that could be implemented (e.g. parallel or lazy rehashing, bucket representation optimization).

The most interesting result to me is the second benchmark where the **kcas** based `Hashtbl` performs worst. 
 I haven't yet analyzed the reasons behind that.

bench/hmap.ml
c-cube: I agree that the hash shouldn't be too expensive, *BUT* a bad hash function (such as `Fun.id`) will also impact the hashtable's performance, by potentially making some buckets a lot larger than others.

In containers I use a reimplementation of FNV ([godbolt here](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:ocaml,selection:(endColumn:25,endLineNumber:18,positionColumn:25,positionLineNumber:18,selectionStartColumn:25,selectionStartLineNumber:18,startColumn:25,startLineNumber:18),source:'let+hash+(n+:+int)+:+int+%3D%0A++let+offset_basis+%3D+0xcbf29ce484222325L+in%0A++let+prime+%3D+0x100000001b3L+in%0A%0A++let+h+%3D+ref+offset_basis+in%0A++for+k+%3D+0+to+7+do%0A++++(h+:%3D+Int64.(mul+!!h+prime))%3B%0A++++(*+h+:%3D+h+xor+(k-th+byte+of+n)+*)%0A++++h+:%3D+Int64.(logxor+!!h+(of_int+((n+lsr+(k+*+8))+land+0xff)))%0A++done%3B%0A++(*+truncate+back+to+int+and+remove+sign+*)%0A++Int64.to_int+!!h+land+max_int%0A%0A%0Alet+()+%3D%0A++let+i+%3D+Sys.opaque_identity+42+in%0A++let+h+%3D+hash+i+in%0A++Printf.printf+%22%25d%5Cn%22+h'),l:'5',n:'0',o:'OCaml+source+%231',t:'0')),k:50,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ocaml5000flambda,deviceViewOpen:'1',filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:ocaml,libs:!(),options:'-O3',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+ocamlopt+5.0.0-flambda+(Editor+%231)',t:'0')),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)) which should not be _too_ expensive.

c-cube:
If it's doable, I'd also be interested in a comparison with a locked structure that would basically resemble:

```ocaml
module Int_tbl = Hashtbl.Make(Int)

type 'a table = {
  shards: 'a Int_tbl.t Rw_lock.t array;
}
```

in the style of rust's [dashmap](https://github.com/xacrimon/dashmap/). The sharding, based e.g. on the lower bits of the hash, should improve behavior under contention.

bench/hmap.ml
polytypic: Note that the benchmark uses random integers as keys.  More generally it is not difficult to avoid bad key sequences in benchmarks.  At a minimum, one should use the same user specified hash for all tables.

lyrm:
Closing this PR as #117 proposed a better implementation of the same algorithm.

PR #64, opened 2023-03-10
Fix space leak of Michael-Scott queue and some tweaks

polytypic:
Fixes issue #64


src/michael_scott_queue.ml
```diff
@@ -40,13 +42,21 @@ let pop q =
     in
     match nhead with
     | Nil -> None
-    | Next (v, _) when Atomic.compare_and_set q.head s nhead -> Some v
+    | Next (v, x) when Atomic.compare_and_set q.head s (Next (Obj.magic (), x))
```
polytypic: Here we allocate a new node for the head.  This way we don't keep a reference to the value `v`.  The same needs to be done in `clean_until`.

An alternative would be to make nodes hold the value in a mutable field and assign `Obj.magic ()` to the field after a successful CAS.  Unfortunately that would break the snapshot functionality, because then, after calling snapshot, pop operations on the queue would destroy the snapshot.

src/michael_scott_queue.ml
```diff
@@ -55,13 +65,14 @@ let push q v =
...
-      find_tail_and_enq n newnode;
-      ignore (Atomic.compare_and_set q.tail tail newnode)
+  let new_tail = Atomic.make Nil in
+  let newnode = Next (v, new_tail) in
+  let old_tail = Atomic.get q.tail in
+  find_tail_and_enq old_tail newnode;
+  if Atomic.compare_and_set q.tail old_tail new_tail then
+    match Atomic.get new_tail with
+    | Nil -> ()
+    | Next (_, new_new_tail) -> fix_tail q.tail new_tail new_new_tail
```
polytypic: If we get to this point, it means that we won the race to update the tail.  However, someone else has already queued another element into the queue.  The protocol chosen here is then that the winner must make sure that the tail is in sync, so we call `fix_tail`, which keeps on updating the tail for as long as it keeps on winning.

polytypic:
BTW, it seems (unsurprisingly) that avoiding false sharing with the MS queue has a very noticeable effect on performance.

The whole idea of the the MS queue is that the head and tail are two distinct shared memory locations and the push and pop operations only access one of those.   Naïvely allocating the head and tail atomics leads directly to high probability of false sharing.

Here is a benchmark run where the implementation avoids false sharing:

```
Hand-written Lockfree.MSQueue: mean = 0.005300, sd = 0.000000 tp=56601255.915217
```

Without false sharing avoidance, the timings from the same benchmark vary highly and are sometimes more than twice as high and basically never as good as above.

To be clear, this PR does not include the change to avoid false sharing:

```diff
 let create () =
   let next = Atomic.make Nil in
-  { head = Atomic.make next; tail = Atomic.make next }
+  Multicore_magic.copy_as_padded
+    {
+      head = Multicore_magic.copy_as_padded @@ Atomic.make next;
+      tail = Multicore_magic.copy_as_padded @@ Atomic.make next;
+    }
```

lyrm:
Thanks, that is a great fix !  I just finished reviewing it and everything looks good.

I will have a look at why the CI checks are failing tomorrow.

polytypic:
> I will have a look at why the CI checks are failing tomorrow.

CI says:
-  ```
    # ocamlfind: Package `threads' not found
    ```

Perhaps related to [this](https://discuss.ocaml.org/t/psa-workaround-when-upgrading-to-ocamlfind-1-9-6-with-ocaml-5-in-opam/11631)?

PR #65, opened 2023-03-10
Lockfree Skip List

sooraj-srini:
Implementation of a lock free skiplist taken directly from the Java implementation given in Herlihy et al's "The Art of Programming" 2nd Edition, section 14.4. There is a different implementation of [contains] from the textbook as the textbook implementation contains typos.

### To do in this PR
- [x] Make it polymorphic 
- [x] Debug : `get_random_level` : returns a level of at least 1, why not 0 ? (it is not working with `0` but why ?)
- [ ]  More documentation
- [ ] Add benchmarks with comparison to naive lock implementation and with the lazy skiplist of #90.
- [ ]  Change `get_random_level` to remove the loop (as @polytypic suggested)

### To do in a future PR
- [ ] a version with hashed key : it would enable optimization on integers and avoid some bad cases related to sorted additions (not sure about that)
- [ ] a version functorized over key type and compare function : key comparison is done with `=` and `<` so with some types, it could be very costly or just not work at all.


Sudha247 (commented):
Thanks for the work @sooraj-srini! This is going in the right direction. I did a quick pass and left some comments below. Will try to do another pass on `add` and `remove` functions.

src/atomicskiplist.ml
```diff
@@ -0,0 +1,242 @@
...
+
+(** create_new_node: creates a new node with some value and height *)
+let create_new_node value height =
+  let next =
+    let arr =
+      Array.make (max_height + 1) (Atomic.make { node = None; marked = false })
+    in
+    for level = 1 to max_height do
+      arr.(level) <- Atomic.make { node = None; marked = false }
+    done;
```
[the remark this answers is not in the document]
sooraj-srini: The reason I used the for loop to initialize the array is so that all the elements of the array would be physically different. A change to one of the elements of the array should not reflect in the other elements. I suppose there might be a more idiomatic way of achieving this.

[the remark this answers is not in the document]
polytypic: `Array.init` ?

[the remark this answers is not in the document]
Sudha247: I see. Yes, `Array.init` should do the trick.

src/atomicskiplist.ml
```diff
@@ -0,0 +1,221 @@
...
+(** create_dummy_node_array: Creates a new array with the different node for each index *)
+let create_dummy_node_array () =
+  let arr = Array.init (max_height + 1) (fun _ -> create_new_node Int.max_int max_height)
+  in
+  arr
+
+(** Get a random level from 1 till max_height (both included) *)
+let get_random_level () =
+  let rec count_level cur_level =
+    if cur_level == max_height || Random.float 1.0 <= 0.5 then cur_level
```
polytypic, line 29: Why not use `Random.bool ()`?

polytypic: BTW, instead of using a loop, one could roughly:
1. Compute `max_height - 1` random bits: `Random.bits () land ((1 lsl (max_height - 1)) - 1) lor (1 lsl (max_height - 1))`.
2. Use the technique described in [this paper](http://supertech.csail.mit.edu/papers/debruijn.pdf) to find the index of the lowest 1 bit.
3. Add 1.

src/atomicskiplist.ml
```diff
@@ -0,0 +1,221 @@
...
+  in
+  let next = Array.init (max_height + 1) (fun _ -> Atomic.make { node = Some tail; marked=false; }) 
+  in
+  let head = { key = Int.min_int; height = max_height; next } in
+  { head }
+
+(** get_mark_ref: Returns the node and the mark from an Atomic markablereference *)
+let get_mark_ref atomic_ref =
+  let ref = Atomic.get atomic_ref in
+  (ref.node, ref.marked)
```
polytypic, line 53: It should be unnecessary to copy the values from the immutable record.  This whole helper function could be just removed.

src/atomicskiplist.ml
```diff
@@ -0,0 +1,221 @@
...
+    Irrespective of return value, fills the preds and succs array with 
+    the predecessors nodes with smaller key and successors nodes with greater than 
+    or equal to key 
+  *)
+let find_in key preds succs sl =
+  let head = sl.head in
+  let init level =
+    let prev = Some head in
+    let curr = get_ref (Option.get prev).next.(level) in
+    let succ, mark = get_mark_ref (Option.get curr).next.(level) in
```
polytypic, line 85: This could be just `let {node = succ; marked = mark} = Atomic.get (Option.get curr).next.(level)`.

Also, `Option.get` should generally be avoided.

src/atomicskiplist.ml
sooraj-srini: The Java implementation of `AtomicMarkableReference` includes functions like `getReference` and `get` which are analogous to `get_ref` and `get_mark_ref` in my implementation. In my aim to stick as close as possible to the implementation in "The Art of Multiprocessor Programming", I used similar functions.

polytypic: I guess that is a reasonable approach to get a working reference implementation.  However, Java and OCaml have many differences that make a direct transliteration from Java to OCaml undesirable.  For example, using options to emulate `null` is inefficient, because in OCaml that adds a level of indirection.  I would not recommend such an approach (i.e. direct transliteration) for a serious implementation.

kayceesrk:
@sooraj-srini will finish his course project in 2 weeks. I wondered what needs to be done to get this across the line to a merge while we have his attention. @Sudha247 @polytypic?

test/atomicskiplist/atomic_skiplist_dscheck.ml
```diff
@@ -0,0 +1,28 @@
+(* This dscheck testcase is not terminating. *)
```
lyrm, line 1: This `dscheck` test takes forever (probably because of too many possible interleavings) but it can actually finish. 

It also displays a weird behavior (if `find` calls are replaced by `add` calls) which is due to randomness (as explained in [this `dscheck` issue](ocaml-multicore/dscheck#20)). It is easily avoided by adding  `Random.init 0` at the beginning of the test (right after `Atomic.trace (fun () ->`).

lyrm (commented):
I did a primary review (I did not dive deep into the algorithm). 

It is a great start : the implementation seems to work. I have done `STM` test (that I will add to the #61). It works. The single `dscheck` test takes a very long time but finished. I will add some and run them through a night to see if any issue raises.

I however think quite a bit of things need to be polished before merging: the code needs to be cleaned and there is improvement to do to better use Ocaml features, as @polytypic mentioned. 

In a general way: 
- formatting (`dune build @fmt` and then `dune promote`)
- documentation in `.mli`

About the code itself, I wrote a few comments, but there seems to be a lot of avoidable copies and indirections. I will try to dive deeper into it at the beginning of next week  to help move it forward quickly.

src/atomicskiplist.ml
```diff
@@ -0,0 +1,215 @@
+type 'a markable_reference = { node : 'a; marked : bool }
+(** markable reference: stores a reference to a node and has a field to specify if it is marked *)
+
+type 'a node = {
+  key : int;
+  height : int;
+  next : 'a node markable_reference Atomic.t array;
+}
+
+type 'a t = { head : 'a node }
```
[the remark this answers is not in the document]
lyrm: (I should add this is probably not the best types for this implementation, just a small improvement of the current ones).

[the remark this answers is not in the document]
lyrm: With the addition of `max_height` as a changeable parameter, `t` can actually be defined as a record. 
 ```ocaml
type t = { head : node; max_height : int} 
```

Also,  it is easy to make this implementation a polymorphic skip list with the use of `Hashtbl.hash` (in which case, `Int.min_int` can be replace by -1) and a node defined as : 

```ocaml
type 'a node = {
  key : int;
  value : 'a;
  height : int;
  next : 'a markable_reference Atomic.t array;
}
```
 
But that can be done once everything else works.

lyrm (commented):
A more complete review of the code. A few important things to do : 
- add `max_height` as a (possibly optional) parameter of `create`.
- remove the `get_mark_ref` and `get_ref` functions
- minor changes in the `add` function to avoid calling `Atomic.get` on a newly created and still unlinked node
- rename `find` to `mem`
- either  correct the types (as it is an integer skip list) or add the necessary changes to make it a polymorphic one.

lyrm:
With the proper changes, this following `dcheck tests` are finishing and passing (with the[ `source-set` dscheck](https://github.com/ocaml-multicore/dscheck/pull/18) branch, that is most likely going to be merged soon). 
```ocaml
open Atomicskiplist

let _two_find () =
  Atomic.trace (fun () ->
      Random.init 0;
      let sl = create ~max_height:2 () in
      let added1 = ref false in
      let found1 = ref false in
      let found2 = ref false in

      Atomic.spawn (fun () ->
          added1 := add sl 1;
          found1 := find sl 1);

      Atomic.spawn (fun () -> found2 := find sl 2);

      Atomic.final (fun () ->
          Atomic.check (fun () -> !added1 && !found1 && not !found2)))

let _two_add () =
  Atomic.trace (fun () ->
      Random.init 0;
      let sl = Atomicskiplist.create ~max_height:2 () in
      let added1 = ref false in
      let added2 = ref false in

      Atomic.spawn (fun () -> added1 := add sl 1);
      Atomic.spawn (fun () -> added2 := add sl 2);

      Atomic.final (fun () ->
          Atomic.check (fun () -> !added1 && !added2 && find sl 1 && find sl 2)))

let _two_add_same () =
  Atomic.trace (fun () ->
      Random.init 0;
      let sl = Atomicskiplist.create ~max_height:2 () in
      let added1 = ref false in
      let added2 = ref false in

      Atomic.spawn (fun () -> added1 := add sl 1);
      Atomic.spawn (fun () -> added2 := add sl 1);

      Atomic.final (fun () ->
          Atomic.check (fun () ->
              (!added1 && not !added2)
              || (((not !added1) && !added2) && find sl 1))))

let _two_remove_same () =
  Atomic.trace (fun () ->
      Random.init 0;
      let sl = create ~max_height:1 () in
      let added1 = ref false in
      let removed1 = ref false in
      let removed2 = ref false in

      Atomic.spawn (fun () ->
          added1 := add sl 1;
          removed1 := remove sl 1);
      Atomic.spawn (fun () -> removed2 := remove sl 1);

      Atomic.final (fun () ->
          Atomic.check (fun () ->
              !added1
              && ((!removed1 && not !removed2) || ((not !removed1) && !removed2))
              && not (find sl 1))))

let _two_remove () =
  Atomic.trace (fun () ->
      Random.init 0;
      let sl = create ~max_height:1 () in
      let added1 = ref false in
      let removed1 = ref false in
      let removed2 = ref false in

      Atomic.spawn (fun () ->
          added1 := add sl 1;
          removed1 := remove sl 1);
      Atomic.spawn (fun () -> removed2 := remove sl 2);

      Atomic.final (fun () ->
          Atomic.check (fun () ->
              let found1 = find sl 1 in
              !added1 && !removed1 && not !removed2 && not found1)))

let () =
  let open Alcotest in
  run "atomic_skiplist_dscheck"
    [
      ( "basic",
        [
          test_case "2-find" `Slow _two_find;
          test_case "2-add-same" `Slow _two_add_same;
          test_case "2-add" `Slow _two_add;
          test_case "2-remove-same" `Slow _two_remove_same;
          test_case "2-remove" `Slow _two_remove;
        ] );
    ]


```

sooraj-srini:
Sure, I have added @lyrm to [my fork](https://github.com/sooraj-srini/lockfree) as a collaborator. 
Additionally, I had written some extra benchmarks in [this repository](https://github.com/sooraj-srini/benchmarks) along with an implementation of a composable skip list using `kcas`. I had hoped to have the time to include these my fork, but did not manage to do so.

lyrm:
(Sorry, some *hmmm* git shenanigans)

lyrm:
I have merged all the small changes/improvements/debugs I did on this implementation. There are still some stuff that can be done, but nothing that should change massively the implementation. I will list the tasks to do (in this PR or in a future one) in the first comment. 

@polytypic  : Could you review this ? I guess at this point, I am mostly interested on optimization/improvement that could be done in this algorithm. In particular, I  think the implementation suffer a lot from false sharing, as the `next` field of a node is an array. This is particularly true for the `head` node that is passed by at every function call. I am not sure on how to improve that.

bench/bench_skiplist.ml
```diff
@@ -0,0 +1,57 @@
...
+  let push () =
+    Domain.spawn (fun () ->
+        let start_time = Unix.gettimeofday () in
+        for i = 0 to (num_elems - 1) / num_threads do
+          Domain.cpu_relax ();
+          let prob = Random.float 1.0 in
+          if prob < add then Skiplist.add sl (Random.int 10000) |> ignore
+          else if prob >= add && prob < add +. remove then
+            Skiplist.remove sl (Random.int 10000) |> ignore
+          else Skiplist.mem sl elems.(i) |> ignore
```
polytypic, line 15: The use of `elems` array here and above is a bit strange.  A huge array is initialized and then only the fraction for a single thread is used.  Why not use a `Random.int` here as well?

bench/bench_skiplist.ml
```diff
@@ -0,0 +1,57 @@
...
+        done;
+        start_time)
+  in
+  let threads = List.init num_threads (fun _ -> push ()) in
+  let start_time_threads =
+    List.map (fun domain -> Domain.join domain) threads
+  in
+  let end_time = Unix.gettimeofday () in
+  let time_diff = end_time -. List.nth start_time_threads 0 in
```
polytypic, line 24: Hmm... The timing collection seems to just use the start time of the first domain and the time after joining all of the threads.

How about:

1. Use a barrier to synchronize all the domains before their loops.
2. Individually time the loop inside each domain and return that from each domain.

Various measures could then be calculated from the collection of timings.  E.g. compute average of the times from each domain to get roughly the same kind of measurement as here, but taking all domains into account rather than the start time of the first domain and the end time (+ some) of the domain that finished last.

src_lockfree/skiplist.ml
```diff
@@ -0,0 +1,193 @@
...
+    if found then false
+    else
+      let new_node_next =
+        Array.map
+          (fun element ->
+            let mark_ref = { node = element; marked = false } in
+            Atomic.make mark_ref)
+          succs
+      in
+      let new_node = { key; height = top_level; next = new_node_next } in
```
polytypic, line 105: The `next` array of the new node is initialized to contain as many elements as the maximum number of levels, but only a part (according to `height`) of those are actually used.

src_lockfree/skiplist.ml
```diff
@@ -0,0 +1,193 @@
...
+  let rec repeat () =
+    let found = find_in (key, preds, succs, sl) in
+    if found then false
+    else
+      let new_node_next =
+        Array.map
+          (fun element ->
+            let mark_ref = { node = element; marked = false } in
+            Atomic.make mark_ref)
+          succs
```
polytypic, line 103: Hmm... Aside from the array created here being too large, I believe there might be a more subtle space leak issue.

What happens here is that a new node is being constructed and as a part of that an array of references to other nodes is created.  A particular reference of this array is not updated during `add` nor are the references in the array created here subject to updates until a reference to the `new_node` is added to a predecessor node _at a specific level_.

Consider the following scenario:

1. A domain performing `add` is suspended after creating the `new_node_next` array.
2. Another domain `remove`s one of the nodes to which the `new_node_next` array has a reference at a non-zero level.
3. The domain performing `add` is resumed and completes the operation.

What will happen then that the `add` will notice that a successor node was removed at around line 118+ as the `compare_and_set_mark_ref` fails.  The `add` will then call `find_in` to update the `preds` and `succs`.  This will not, however, update the reference in the `new_node_next` array, which was created based on an earlier `succs`.  That is because the reference is at a level on which the `new_node` is not yet attached to the skip list (that is because the `compare_and_set_mark_ref` failed).

This means that after `add` returns, the `new_node` has been added to the skip list and the `new_node` contains a reference to a removed node.  This means that the `key` contained in that removed node cannot be garbage collected.  It will remain in memore until some call to `find_in` will notice that the removed node (due to marked references) and removes it.  But there is no guarantee such a call will happen. It might never happen.

Am I missing something?

polytypic:
I wrote a lock-free skiplist from scratch somewhat inspired by what I learned from reviewing the code in this PR.  I used a number of techniques to optimize it and it is roughly 1.75 times faster (6.34 M op/s vs 3.55 M op/s) on the benchmark in this PR.  You can find the code in this [gist](https://gist.github.com/polytypic/70d5f7981a61a8c4ad4b9cce590d6f72).  The code in the gist has some comments on some of the optimizations.  The main improvement likely comes from the internal representation that avoids a level of indirection and takes less memory.

kayceesrk:
I don't see a reason to retain this PR if the new implementation is 1.75x faster. It may be best to close this PR and open a new one with the code from the gist.

lyrm:
I'm closing this PR as an improved implementation of this skiplist algorithm is proposed in PR #99. Thanks to all contributors!

PR #68, opened 2023-03-22
Rendezvous and demonstration on Spsc_queue API

art-w:
This is a small proposal to avoid spinlocks in this repository by introducing some mean of synchronizing domains. I think it would help users if the APIs provided facilities for this common pattern, while working well with fibers and custom schedulers (so not just system locks in place of spinlocks).

I did some experiment on the `Spsc_queue` benchmark where a spinlock is used for the consumers/producers when the queue is empty/full. Currently this is absurdly slow if the two domains happen to be scheduled on the same CPU core: it takes up to 3s to push/pop a thousand elements, rather than milliseconds on two cores! (.. this is why the CI benchmarks don't terminate in any reasonnable time atm, you can try it with `taskset --cpu-list 1 ./_build/default/bench/main.exe` but I recommend lowering the items counts first)

Adding a couple of external semaphores to this benchmark solves the single core worst-case for spinlocks, but at the price of being 4x slower in general. By integrating the locking logic into the datastructure, we can achieve nearly no observable regression in performances while also providing an API that don't encourage users to spinlock. It's debatable if the resulting datastructure is still "lockfree", but at least the locking is opt-in :-°

This PR builds on [a previous experiment in domainslib](https://github.com/ocaml-multicore/domainslib/pull/98) where the goal was to synchronize fibers with an `Mpmc_queue`, by using a "rendezvous" lock between matching pop and push operations. Essentially we want some mean for a `push`ing domain to unlock a stuck `pop`ing domain (which happens when the queue is empty, but the user has signified that it really wants to wait). Dually if the queue was full, we want the next `pop` to unstuck a waiting `push`ing domain. By exploiting the custom scheduler, locking is cheap since the domain can keep working on other tasks.

... Now this is just a quick demo to gather feedback on the proposed `Rendezvous` abstraction, there's still a bunch of spinlocks to remove from this repo if there's interest. Some questions:

- This PR example only uses a `unit Rendezvous.t`, so perhaps the rendezvous polymorphism is not required and could be added on the side on a case-by-case basis? I remember that it was nice for the `Mpmc_queue` in domainslib, and the type is a bit easier to read than with `unit` everywhere (?)
- I tried a rendezvous type `(('a -> unit) -> unit) -> 'a` since @polytypic used something similar in kcas, but it was less convenient to handle the edge-case "I thought I needed to lock but don't anymore"... The callback pair is less fancy but easier to understand, maybe (?)
- Should the `Rendezvous` type definition be public? I did so with the hope that other datastructure libraries could adopt the pattern and custom schedulers would be able to provide implementations, without requiring an explicit dependency on `lockfree` just for this type, but perhaps that won't be the case... We could expose standard synchronizations methods like mutexes instead then.

### DIFF
--- a/src/spsc_queue.ml
+++ b/src/spsc_queue.ml
@@ -22,10 +22,11 @@
  *)
 
 type 'a t = {
-  array : 'a Option.t Array.t;
+  array : 'a option array;
   tail : int Atomic.t;
   head : int Atomic.t;
   mask : int;
+  lock : (unit -> unit) option Atomic.t;
 }
 
 exception Full
@@ -36,22 +37,31 @@ let create ~size_exponent =
     head = Atomic.make 0;
     tail = Atomic.make 0;
     mask = size - 1;
-    array = Array.init size (fun _ -> None);
+    array = Array.make size None;
+    lock = Atomic.make None;
   }
 
-let push { array; head; tail; mask; _ } element =
+let unlock lock =
+  if Atomic.get lock <> None then
+    match Atomic.exchange lock None with
+    | Some release -> release ()
+    | None -> ()
+
+let try_push { array; head; tail; mask; lock } element =
   let size = mask + 1 in
-  let head_val = Atomic.get head in
   let tail_val = Atomic.get tail in
-  if head_val + size == tail_val then raise Full
+  let head_val = Atomic.get head in
+  if head_val + size = tail_val then false
   else (
     Array.set array (tail_val land mask) (Some element);
-    Atomic.set tail (tail_val + 1))
+    Atomic.set tail (tail_val + 1);
+    unlock lock;
+    true)
 
-let pop { array; head; tail; mask; _ } =
+let try_pop { array; head; tail; mask; lock } =
   let head_val = Atomic.get head in
   let tail_val = Atomic.get tail in
-  if head_val == tail_val then None
+  if head_val = tail_val then None
   else
     let index = head_val land mask in
     let v = Array.get array index in
@@ -59,6 +69,29 @@ let pop { array; head; tail; mask; _ } =
     Array.set array index None;
     Atomic.set head (head_val + 1);
     assert (Option.is_some v);
+    unlock lock;
     v
 
 let size { head; tail; _ } = Atomic.get tail - Atomic.get head
+
+let wait ~rdv t expected_size =
+  let release, wait = rdv () in
+  let some_release = Some release in
+  if Atomic.compare_and_set t.lock None some_release then
+    if size t = expected_size then wait ()
+    else
+      let (_ : bool) = Atomic.compare_and_set t.lock some_release None in
+      ()
+  else unlock t.lock
+
+let rec push ~rdv t element =
+  if not (try_push t element) then (
+    wait ~rdv t (t.mask + 1);
+    push ~rdv t element)
+
+let rec pop ~rdv t =
+  match try_pop t with
+  | Some v -> v
+  | None ->
+      wait ~rdv t 0;
+      pop ~rdv t
--- a/bench/bench_spsc_queue.ml
+++ b/bench/bench_spsc_queue.ml
@@ -1,24 +1,20 @@
 open Lockfree
 
 let item_count = 2_000_000
-
-let rec try_until_success f =
-  try f () with Spsc_queue.Full -> try_until_success f
+let rdv = Rendezvous.semaphore_unit
 
 let run () =
-  let queue = Spsc_queue.create ~size_exponent:3 in
+  let queue = Spsc_queue.create ~size_exponent:10 in
   let pusher =
     Domain.spawn (fun () ->
         let start_time = Unix.gettimeofday () in
         for i = 1 to item_count do
-          try_until_success (fun () -> Spsc_queue.push queue i)
+          Spsc_queue.push ~rdv queue i
         done;
         start_time)
   in
-  for _ = 1 to item_count do
-    while Option.is_none (Spsc_queue.pop queue) do
-      ()
-    done
+  for i = 1 to item_count do
+    assert (Spsc_queue.pop ~rdv queue = i)
   done;
   let end_time = Unix.gettimeofday () in
   let start_time = Domain.join pusher in
--- a/src/lockfree.ml
+++ b/src/lockfree.ml
@@ -25,6 +25,7 @@
 Copyright (c) 2017, Nicolas ASSOUAD <nicolas.assouad@ens.fr>
 ########
 *)
+module Rendezvous = Rendezvous
 module Ws_deque = Ws_deque
 module Spsc_queue = Spsc_queue
 module Mpsc_queue = Mpsc_queue
--- a/src/lockfree.mli
+++ b/src/lockfree.mli
@@ -30,6 +30,7 @@ Copyright (c) 2017, Nicolas ASSOUAD <nicolas.assouad@ens.fr>
 ########
 *)
 
+module Rendezvous = Rendezvous
 module Ws_deque = Ws_deque
 module Spsc_queue = Spsc_queue
 module Mpsc_queue = Mpsc_queue
--- a/src/rendezvous.ml
+++ b/src/rendezvous.ml
@@ -1,0 +1,67 @@
+type 'a t = unit -> ('a -> unit) * (unit -> 'a)
+
+let fail () = failwith "Rendezvous: double send"
+
+let semaphore () =
+  let r = ref None in
+  let s = Semaphore.Binary.make false in
+  let send v =
+    r := Some v;
+    Semaphore.Binary.release s
+  and recv () =
+    Semaphore.Binary.acquire s;
+    Option.get !r
+  in
+  (send, recv)
+
+let semaphore_unit () =
+  let s = Semaphore.Binary.make false in
+  let send () = Semaphore.Binary.release s
+  and recv () = Semaphore.Binary.acquire s in
+  (send, recv)
+
+let backoff ?min_wait ?max_wait () =
+  let r = Atomic.make None in
+  let backoff = Backoff.create ?min_wait ?max_wait () in
+  let rec send v = if not (Atomic.compare_and_set r None (Some v)) then fail ()
+  and recv () =
+    match Atomic.get r with
+    | Some v -> v
+    | None ->
+        Backoff.once backoff;
+        recv ()
+  in
+  (send, recv)
+
+let backoff_unit ?min_wait ?max_wait () =
+  let r = Atomic.make false in
+  let backoff = Backoff.create ?min_wait ?max_wait () in
+  let rec send () = if not (Atomic.compare_and_set r false true) then fail ()
+  and recv () =
+    if not (Atomic.get r) then (
+      Backoff.once backoff;
+      recv ())
+  in
+  (send, recv)
+
+let spinlock () =
+  let r = Atomic.make None in
+  let rec send v = if not (Atomic.compare_and_set r None (Some v)) then fail ()
+  and recv () =
+    match Atomic.get r with
+    | Some v -> v
+    | None ->
+        Domain.cpu_relax ();
+        recv ()
+  in
+  (send, recv)
+
+let spinlock_unit () =
+  let r = Atomic.make false in
+  let rec send () = if not (Atomic.compare_and_set r false true) then fail ()
+  and recv () =
+    if not (Atomic.get r) then (
+      Domain.cpu_relax ();
+      recv ())
+  in
+  (send, recv)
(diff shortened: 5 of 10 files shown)


polytypic:
A quick note: I currently strongly feel this sort of thing should be put into a minimal library of its own.  IOW, I would not put it into `lockfree` (nor [`kcas`](https://github.com/ocaml-multicore/kcas/pull/32) for that matter).  The ability to suspend and later resume seems to be useful/needed for a lot of low level facilities and would likely be implemented by any and all libraries providing schedulers of some sort that want to co-operate.

src/spsc_queue.ml
```diff
@@ -36,29 +37,61 @@ let create ~size_exponent =
...
+
+let wait ~rdv t expected_size =
+  let release, wait = rdv () in
+  let some_release = Some release in
+  if Atomic.compare_and_set t.lock None some_release then
+    if size t = expected_size then wait ()
+    else
+      let (_ : bool) = Atomic.compare_and_set t.lock some_release None in
+      ()
+  else unlock t.lock
```
polytypic, line 85: Hmm... Did I understand correctly that if two calls to `wait` are being made, e.g. to `pop`s when the queue is empty, then what happens is that the second call to `wait` actually wakes up the first one?  So, in that case you get a very expensive busy wait?

src/spsc_queue.ml
```diff
@@ -36,29 +37,61 @@ let create ~size_exponent =
...
 
 let size { head; tail; _ } = Atomic.get tail - Atomic.get head
+
+let wait ~rdv t expected_size =
+  let release, wait = rdv () in
+  let some_release = Some release in
+  if Atomic.compare_and_set t.lock None some_release then
+    if size t = expected_size then wait ()
+    else
+      let (_ : bool) = Atomic.compare_and_set t.lock some_release None in
```
polytypic, line 83: Hmm... I wonder whether this can cause problems?  It is possible that another party either manages to change the size of the queue (before the first CAS) or call `release` (between the two CASes).

Related to this, to avoid capturing the stack unnecessarily (before the waiters are added), the kcas blocking draft uses [three states](https://github.com/ocaml-multicore/kcas/blob/1d03f71bba77b920b38049f4f78db62e4ad4f36a/src/kcas.ml#L40): `Init | Resumed | Waiting of ...`.  Then both [suspend](https://github.com/ocaml-multicore/kcas/blob/1d03f71bba77b920b38049f4f78db62e4ad4f36a/src/kcas.ml#L638-L651) and [resume](https://github.com/ocaml-multicore/kcas/blob/1d03f71bba77b920b38049f4f78db62e4ad4f36a/src/kcas.ml#L547-L549) take the race condition (where resume happens before suspend is fully ready) into account.

polytypic:
> perhaps the rendezvous polymorphism is not required and could be added on the side on a case-by-case basis?

It is difficult to say, but generally speaking, there are likely to be a lot of cases where the ability to pass value is not needed (only the suspend/resume is needed), and it usually requires a bit more work (an allocation or two) to make it possible to pass a value.

src/spsc_queue.ml
```diff
@@ -36,29 +37,61 @@ let create ~size_exponent =
...
+      ()
+  else unlock t.lock
+
+let rec push ~rdv t element =
+  if not (try_push t element) then (
+    wait ~rdv t (t.mask + 1);
+    push ~rdv t element)
+
+let rec pop ~rdv t =
```
polytypic, line 92: I also used a(n optional) named parameter to pass in the suspend/resume mechanism in my kcas blocking draft.

However, what I was thinking is that the suspend/resume (or rendezvous) mechanism could be something that can be obtained from the DLS, for example, at the point when it is needed.

A hypothetical rendezvous framework would provide a way to install a rendezvous mechanism:

```ocaml
type t = (* ((unit -> unit) -> unit) -> unit   OR   whatever  *)

val set : t -> unit
val get : unit -> t (* Raises or returns default mechanism if nothing has been explicit set *)
```

Schedulers, like Eio, would call `set` to install the mechanism when they run their event loops.  (Perhaps something like `push` and `pop` rather than `set` might be a better API.)

Facilities that need the ability then call `get` when they need it.

This would allow things like kcas to work independently of the scheduler(s).  You could have domainslib running on some domains and Eio on others.  And you could have communication between the domains &mdash; without having to know which mechanism to pass from which domain.

src/spsc_queue.ml
polytypic: Ah.. SPSC! Right.

src/spsc_queue.ml
polytypic: I guess SPSC makes things a bit simpler here as well.

src/spsc_queue.ml
art-w: Yeah :P This was a quick example that only works for SPSC, but in general multiple waiting `pop` would actually push their "release" function onto a queue. Then the next `push` would see the value queue as empty and pop/wake-up the first available `pop`.

src/spsc_queue.ml
kayceesrk: @Sudha247 is also involved in this effort with @deepali2806. We're using suspend effect now to safely use lazy from different threads. Think blackholing for thunk evaluation as in GHC.

src/spsc_queue.ml
art-w: Yes it's possible for another party to call `release` even though `wait ()` will not be called... I think it's fine? (however, we do need to be careful when calling `wait ()` otherwise it's possible for the queue to be unstuck just before we install our lock, followed by no more operations to wake us up)

(I believe the `size t = expected_size` is doing something similar to your three states to determine the `Resumed` state implicitly for the queue internals)

art-w:
> I currently strongly feel this sort of thing should be put into a minimal library of its own.

Yeah I don't have an opinion on this.. It really depends on whether custom schedulers and datastructure libraries need a lot more than an alias for the type `unit -> ('a -> unit) * (unit -> 'a)` (well, datastructures will need some default implementations of this type when testing, but schedulers will not)

> it usually requires a bit more work (an allocation or two) to make it possible to pass a value.

Indeed, that's why I added some specialized `_unit` to show that it shouldn't be too much of an issue (but either way, I'm not seeing a performance difference in my tests, as we only allocate extra stuff when we have to block)

---

To quote your [example from kcas](https://github.com/ocaml-multicore/kcas/blob/1d03f71bba77b920b38049f4f78db62e4ad4f36a/src/kcas.ml#L638-L651) on continuations vs products for the rendezvous type:

```ocaml
scheduler (fun resume ->
    if not (Atomic.compare_and_set self `Init (`Waiting resume))
    then resume ());
```

vs

```ocaml
let resume, wait = scheduler () in
if Atomic.compare_and_set self `Init (`Waiting resume)
then wait ()
```

A small issue I had with the continuation is that you need to call `resume ()` in the happy case when you don't need to lock (so it's slightly awkward for the `scheduler` which now need to check if blocking is necessary or not after the callback completes)

src/spsc_queue.ml
art-w: (To be clear, I totally agree that calling `release` when `wait` might not be called feels dirty when compared to your two-phases commit.. but I don't have an intuition for why this might break custom schedulers assumptions)

polytypic:
> > I currently strongly feel this sort of thing should be put into a minimal library of its own.
> 
> Yeah I don't have an opinion on this..

I'm thinking this more in terms of having the capability available without explicit parameterization.  One approach would be to use the DLS as I suggested previously.  This way one could just use libraries like kcas with blocking &mdash; no matter how many different schedulers are being used in the program things would just work (as long as the schedulers just install the mechanism to DLS when running their event loops).  Passing things through parameters or functorizing everything easily becomes a mess &mdash; althrough it definitely has the advantage of not using ambient state.  Also, if every library would have slightly different definitions (one would use rendezvous, another would use the CPS suspend, third would use a `Suspend` effect), then it would become much less practical to e.g. use DLS.

> A small issue I had with the continuation is that you need to call `resume ()` in the happy case when you don't need to lock (so it's slightly awkward for the `scheduler` which now need to check if blocking is necessary or not after the callback completes)

Hmm... Yes, this could be an advantage of breaking the suspend/resume protocol into two like this.  So, IIUC, basically, in an effects based implementation it would be the `wait` implementation that actually performs the effect to capture the stack.  By making `wait` a simple first-order function, things potentially become simpler.

src/spsc_queue.ml
polytypic: > in `domainslib`, we need the `pool` argument to wait so I'm not sure we can expect schedulers to preconfigure the `dls_rendezvous_key` for the user in general

These sorts of schedulers, including `domainslib` and Eio (and every other similar effects based library I know of for OCaml), basically run a loop of some kind on the domains on which they run.  I don't see why it would be problematic to install the rendezvous mechanism to DLS at the start of the loop.

It would also be fairly easy to provide a function, in `domainslib`, like `get_pool: unit -> pool ` (or `get_pool: unit -> pool option` if you prefer) that obtains the pool with which the current domain is associated with.

One possibility is also to allow both: passing an optional parameter explicit to tell how to block and, in the absence of such parameter, obtain the mechanism from DLS.

art-w: Yeah I don't know why `domainslib` provides the ability to ping-pong between different domain pools... Yet another generic alternative might be to release the blocking domain into a shared pool like [`idle-domains`](https://github.com/polytypic/idle-domains) and expect custom schedulers to integrate with that?

deepali2806: > I unfortunately haven't had time to think about this thoroughly, but I'm not immediately convinced that the features like `resume_result` and returning a `bool` are the best way to go. Are they necessary for best results or could the same be communicated through side-channels efficiently enough? How often are the features needed? (In the `unified_interface` repo the `Resume_failure` constructor does not seem to used at all.) They might be necessary, but I need to think about this. Also, are they sufficient? Is there some special case that would benefit from other special features? I assume these questions have been considered and I'd love to hear some more rationale behind the choices

The return type of resumer captures the notion of cancellation of the task. It means if the particular scheduler cancels the task while it is still in the suspended state, we should avoid resuming it. The resumer is defined in such a way that if the task is live, it actually resumes the task and returns `Resume_success` when the corresponding resume function is called. In the alternate case, when the task is cancelled, it will return `Resume_failure` instead of actually resuming the task. In MVar implementation [here](https://github.com/deepali2806/unified_interface/blob/0b4955b6a1729fb933eb5fdc4562bbaf0737c353/MVar.ml#L47), we are checking the resumer return value. When the resumer return value is `Resume_failure` it means we can skip through it and retry the operation again to get the next resumer.

Also, in the type signature of Suspend effect, we have a function that takes a resumer and returns a `bool` type. It signifies the thread safety while suspending the task. Thread safety, in this case, is achieved through lock-free [implementation](https://github.com/deepali2806/unified_interface/blob/0b4955b6a1729fb933eb5fdc4562bbaf0737c353/MVar.ml#L66). It means we are using atomic operations like compare and swap (CAS). When such a CAS operation is successful, we will return the `true` value, indicating the push to the suspended queue is successful. In case of CAS failure, we have to retry the operation again to get the most recent state of the queue or MVar.

ElectreAAS:
@art-w you scared me, but the benchmarks are run properly and the graphs are present! [Link here](https://autumn.ocamllabs.io/ocaml-multicore/lockfree/pull/68/benchmark/lockfree?worker=autumn&image=ocaml%2Fopam%3Adebian-ocaml-5.0). The problem is that these benchmarks have the name "Lockfree" and hence aren't shown when just clicking on the PR, you also need to click on Lockfree here.
image
However the graphs aren't particularly useful as there is nothing to compare to, but they're present!

polytypic:
Also mentioning here that the [domain-local-await](https://github.com/ocaml-multicore/domain-local-await) or DLA library is now available as an [opam package](https://opam.ocaml.org/packages/domain-local-await/) and provides a blocking mechanism inspired by the "rendezvous" primitive described in this PR.  I plan to soon release a version of [kcas](https://github.com/ocaml-multicore/kcas/) using DLA and there are PRs to add DLA support to [Eio](https://github.com/ocaml-multicore/eio/pull/494) and [Domainslib](https://github.com/ocaml-multicore/domainslib/pull/107).  As described in the [DLA library README](https://github.com/ocaml-multicore/domain-local-await#readme) it may be replaced later by an official blocking mechanism, but I hope that we can use it now to experiment providing blocking abstractions that from the POV of casual users  _just work_ ™ with OCaml 5 today.

lyrm:
@art-w : as we are not moving to blocking data structures for now, can we close this PR ?

art-w:
Yes of course! The ideas have been integrated in DLA and Picos now, this PR is obsolete :)

PR #69, opened 2023-03-23
Set QCHECK_MSG_INTERVAL to avoid clutter in CI logs

jmid:
QCheck by default prints updates to stdout often to give the user a sense of progress.
This works well in a terminal/console, but less well in an append-only CI-log where the many printed lines adds clutter.
By setting the environment variable `QCHECK_MSG_INTERVAL` to something more sensible, we obtain cleaner CI logs.

(We also use this in `multicoretests` and `domainslib` - but it must have been an oversight in `lockfree`)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,9 @@ jobs:
 
     runs-on: ${{ matrix.os }}
 
+    env:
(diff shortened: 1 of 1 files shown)


lyrm:
Thanks !

PR #70, opened 2023-04-03
Mark alcotest as a test dependency

Khady:
from what I could see in the repo, alcotest is only used in tests. So alcotest should be tagged with `with-test` in the opam file.


Sudha247 (approved):
Indeed, thanks!

PR #72, opened 2023-05-11
CI clean up

lyrm:
- cleaning up github actions to remove the ones overlapping with Ocaml CI. 
- add filter to .opam file to avoid non working architectures
- adding Windows in github action

@Sudha247 : not sure of what I did for Windows. Could you have a look ?

Also, how can I check it is working the way I expect it to work ?


Sudha247:
The windows run looks fine to me, but `qcheck-core` has build failures on windows. Sorry to trouble you @jmid, do you happen to know what's up?

Sudha247:
Thanks @jmid! Taking inspiration from multicoretests' setup is perhaps a good place to start, @lyrm.

shym:
Two differences with our workflows come to mind:
- we use the `https://github.com/ocaml-opam/opam-repository-mingw.git#sunset` branch rather than `opam2`,
- we do not use the `duniverse`.
But I admit the error message, saying that `bytes` is missing when it was explicitly installed just before is weird. We have sometimes resorted to setting `OPAMJOBS` to 1 on Windows (mostly for failures that seemed due to memory exhaustion, but it might be a simple workaround here if it is a terrible race condition somewhere).

.github/workflows/main.yml
```diff
@@ -34,15 +25,20 @@ jobs:
...
+          opam-depext: false
+          ocaml-compiler: ocaml.5.0.0,ocaml-option-mingw
           opam-repositories: |
-            default: https://github.com/ocaml/opam-repository.git
-            alpha: https://github.com/kit-ty-kate/opam-alpha-repository.git
+            local: ${{ github.workspace }}\.github\opam\custom
+            dra27: https://github.com/dra27/opam-repository.git#windows-5.0
+            override: https://github.com/shym/custom-opam-repository.git
+            default: https://github.com/ocaml-opam/opam-repository-mingw.git#sunset
+            upstream: https://github.com/ocaml/opam-repository.git
```
jmid, line 40: I don't think you need:
- `local` - as that refers to custom opam files in [the `multicoretests` repo](https://github.com/ocaml-multicore/multicoretests/tree/main/.github/opam) that you don't have here
- `override` - as that refers to @shym's [repo of packages patched for 5.1 and `trunk`](https://github.com/shym/custom-opam-repository/tree/master/packages) which is not tested here

A quick test without those two lines should tell us... :crossed_fingers:

lyrm: Thanks !

lyrm:
Ocaml-ci is now failing due to :
- time out with `dscheck` tests
- out of memory issue with the `lint-lower-bounds` build. 

`dscheck` issue is going to get fixed by a soon-to-come new `dscheck` release. For the second issue, I still need to investigate but I will take care of it later (see #73).

PR #76, opened 2023-07-05
Refactoring to have a lockfree package

lyrm:
(Based on #67)
This PR aims to reorganize the repository to make two packages and prepare for the addition of lock-based and blocking data-structures. 
- `saturn_lockfree` : contains only `lockfree` data structures ;
- `saturn` : contains the "default" data structures. Right now, it is mostly the ones in `saturn_lockfree` except from a sub-module of `Relaxed_queue`. However, in some future, we could have a `lock-based` implementation that we would recommend to use by default over the lock-free one.

### Questions
1. I named the package `Saturn_lockfree` but the library itself is called `lockfree`. Is this ok ?
2. With this refactoring, we get what people have be asking for : a `lockfree` package that has only its own dependencies. I don't see any drawbacks, but I may be missing something. Any comments about that ?


saturn.opam
```diff
@@ -0,0 +1,24 @@
...
+depends: [
+  "ocaml" {>= "4.12"}
+  "dune" {>= "3.0"}
+  "domain_shims" {>= "0.1.0"}
+  "qcheck" {with-test & >= "0.18.1"}
+  "qcheck-stm" {with-test & >= "0.2"}
+  "qcheck-alcotest" {with-test & >= "0.18.1"}
+  "alcotest" {with-test & >= "1.6.0"}
+  "yojson" {>= "2.0.2"}
+  "dscheck" {with-test & >= "0.1.0"}
```
Sudha247, line 20: We should make these test-only dependencies, I think.

PR #78, opened 2023-07-12
added michael_scott bounded queue

DeltaCube23:
Implementation of a bounded version of michael-scott 2-lock queue taken from section 10.3 in art of multiprocessor programming. It provides a push function that only adds elements to the queue if it is not in full capacity, and a pop function that adds waits till the queue is not empty to remove an element from the head.

**TODO**
1) pass capacity of queue as a paramater to init function?
2) split the counter into 2 separate ones each for push and pop, and only synchronize when it reaches full capacity


lyrm:
Thanks !

> **TODO**
> 
>     1. pass capacity of queue as a paramater to init function?
Yes, you can define a reasonable default value and have an optional parameter. Your function with them have a type : 
`val create : ?capacity:int -> unit -> 'a t` 


>     2. split the counter into 2 separate ones each for push and pop, and only synchronize when it reaches full capacity

So when `push_size` reaches `capacity`, it will be updated to `push_size - pop_size`, right ? That seems like a good idea.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+    | Next (value, next) ->
+        t.head <- next;
+        value
+  in
+  let full = Atomic.fetch_and_add t.size (-1) = t.capacity in
+  Mutex.unlock t.head_lock;
+  (* It it was full then signal the waiting push threads to wake up *)
+  if full then (
+    Mutex.lock t.tail_lock;
+    Condition.broadcast t.not_full;
```
polytypic, line 73: Is it necessary to wake up all the push threads?

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
+(*
+  structure node t value: data type, next: pointer to node t
+  structure queue t Head: pointer to node t, Tail: pointer to node t, H lock: lock type, T lock: lock type
+*)
+
+type 'a node = Nil | Next of 'a * 'a node ref
```
polytypic, line 6: You could use a mutable record here:

```ocaml
type 'a node = Nil | Next of { mutable value: 'a; mutable next: 'a node ref }
```

The advantage of this is that the size of a node is now smaller.  This can improve performance (less frequent GC needed, fewer words written to memory, GC examines fewer words).  The disadvantage is that the implemention (avoiding space leaks) is a bit a more intricate as you need to point to a node and must then clear the `value` field after reading it in `pop`.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+
+type 'a t = {
+  mutable head : 'a node ref;
+  mutable tail : 'a node ref;
```
polytypic, line 10: The `head` and `tail` fields are mutated by `pop` and `push` operations, respectively.  Now that both fields are next to each other, it means there is very high likelyhood of [false sharing](https://en.wikipedia.org/wiki/False_sharing), which will degrade performance.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+  head_lock : Mutex.t;
+  tail_lock : Mutex.t;
+  not_empty : Condition.t;
+  not_full : Condition.t;
+  size : int Atomic.t;
```
polytypic, line 15: As the size field is read and written to by every `push` and `pop` it becomes a sequential bottleneck.

I wonder whether it would be possible to adapt the technique I described [here](https://github.com/roburio/miou/issues/2#issuecomment-1630641505) for a lock-free MS queue to get less contentious length counting for the lock-based version.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+  capacity : int;
```
polytypic, line 16: There are use cases for being able to (read and) change the capacity of a bounded queue.  For example, setting the capacity to 0 effectively closes the queue.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+  let dummy = ref Nil in
+  {
+    head = dummy;
+    tail = dummy;
+    head_lock = Mutex.create ();
+    tail_lock = Mutex.create ();
+    not_empty = Condition.create ();
+    not_full = Condition.create ();
+    size = Atomic.make 0;
+    capacity = 1_000_000;
```
polytypic, line 30: Something like `Int.max_int` could be used as a default value to mean that the queue has unlimited capacity (memory will run out before).  It is then also possible to make it so that with the unlimited capacity value the capacity checks are skipped.  With a scalable length tracking this could make the bounded versions (nearly) as fast as an unbounde version.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+  (* Check if queue is empty before this insertion *)
+  t.tail := new_node;
+  t.tail <- new_tail;
+  let empty = Atomic.fetch_and_add t.size 1 = 0 in
+  Mutex.unlock t.tail_lock;
+  (* If it was empty then signal the waiting pop threads to wake up *)
+  if empty then (
+    Mutex.lock t.head_lock;
+    Condition.broadcast t.not_empty;
+    Mutex.unlock t.head_lock)
```
polytypic, line 51: This can be optimized to:

```ocaml
    Mutex.lock t.head_lock;
    Mutex.unlock t.head_lock;
    Condition.broadcast t.not_empty (* or signal *)
```

Why does this work and why is the `Mutex.lock - unlock` even needed?

The key is that a condition variable has no memory, so we must ensure that any potential waiter has called `Condition.wait` before we call `Condition.broadcast` (or `Condition.signal`) or reads the updated (non-zero) `t.size`.

Take a look at the `wait` in `pop`:

```ocaml
  Mutex.lock t.head_lock;
  (* Ensures queue is not empty *)
  while Atomic.get t.size = 0 do
    Condition.wait t.not_empty t.head_lock
  done;
```

It is possible that, if the signaler doesn't take the mutex, then the signal could happen after `Atomic.get` and before `Condition.wait` (and the `wait` might never return).  However, if the signaler takes the lock, i.e. calls `Mutex.lock t.head_lock`, then it must be that either the `Mutex.lock` returns after the `Condition.wait` or it returns before the `Mutex.lock` of a potential `Condition.wait` call.  In either case things will work out.  In first case the `Condition.broadcast` will wake up the waiter.  In the second case the `Atomic.get` before wait will return a non-zero value.

polytypic:
Overall this looks good!

I added some minor comments on how to potentially improve.

src/bounded_queue.ml
```diff
@@ -0,0 +1,113 @@
...
+  let new_node = Next (value, new_tail) in
+  Mutex.lock t.tail_lock;
+  (* Ensures queue is not full *)
+  while Atomic.get t.size = t.capacity do
+    Condition.wait t.not_full t.tail_lock
+  done;
+  (* Check if queue is empty before this insertion *)
+  t.tail := new_node;
+  t.tail <- new_tail;
+  let empty = Atomic.fetch_and_add t.size 1 = 0 in
```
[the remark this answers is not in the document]
DeltaCube23: okay sure, the equality check at line 68 too could be moved outside critical section since similar logic right, or is there any reason why you didn't comment on that.

src/bounded_queue.ml
DeltaCube23: if you don't wake up all respective operation (push/pop) threads then it could result in lost wakeup problem. it's described in section 8.2.2 in art of multiprocessor programming. Essentially one of the threads could miss the wakeup call and need to wait forever. Even just comparing performance in benchmarks using broadcast seems to do better than signal by a slight margin. (Note: the program won't terminate in the scenarios when the lost wakeup problem occurs).

src/bounded_queue.ml
DeltaCube23: done

src/bounded_queue.ml
DeltaCube23: what about the empty case, that will still have to be checked even for this default value

src/bounded_queue.ml
DeltaCube23: what would the function signature for such a method look like (val modify :  'a t -> int -> unit)? I assume you would need to hold both locks in this case. also the check for pushing would need to modified to t.size >= t.capacity correct?

src/bounded_queue.ml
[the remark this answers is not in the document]
polytypic: > equality check at line 68 too could be moved outside critical section

Yes, exactly.

> any reason why you didn't comment on that

I don't want to repeat and I think it is good a practise, as a developer, to look for more instances of any observation.

src/bounded_queue.ml
polytypic: Yes, with "capacity checks" I was referring to needing to check that the capacity of the queue is not exhausted (i.e. not_full).

src/bounded_queue.ml
polytypic: Yes, something like

```ocaml
val set_capacity : 'a t -> int -> unit
```

And, yes, the logic would then be that when `size < capacity` there is room in the data structure.  Otherwise adding more elements should not be allowed, but removing elements should still work.

src/bounded_queue.ml
polytypic: Hmm... I don't have that book and the chapter seems to discuss the topic in general.  Yes, it is possible to use mutexes and conditions in such a way that broadcast is necessary (even if only wait can actually proceed).  However, I don't immediately see how that would happen in this particular case.  Each of the two conditions is `wait`ed for only one piece of state and the state being checked is exactly the one modified prior to corresponding `broadcast`.  I must be missing something.  Could you describe a sequence of operations that causes a lost wakeup?

polytypic: Thinking about this now there is a kind of future issue with this implementation: cancellation.  This doesn't apply yet, because the code uses the Stdlib `Mutex` and `Condition`, which do not support cancellation and are also otherwise scheduler unfriendly (we really shouldn't use those).

Concurrent data structures should ideally be scheduler friendly such that they do not block the thread / domain from running other fibers.  Data structures should also be safe with respect to cancellation such that they do not leave the data structure (or other fibers operating on the data structure) in an invalid state after cancellation.

Assume, for a moment, that Mutex and Condition would actually support cancellation (such as the Mutex and Condition in Eio), and consider the following (code from `pop` with additional comments):

```ocaml
  (* The fiber has been cancelled at this point. *)
  
  let full = Atomic.fetch_and_add t.size (-1) in
  Mutex.unlock t.head_lock;
  
  (* It it was full then signal the waiting push threads to wake up *)
  
  if full = t.capacity then (
    Mutex.lock t.tail_lock;  (* <- As fiber was cancelled, this will raise the cancellation exception... *)
    Mutex.unlock t.tail_lock;
    Condition.broadcast t.not_full);  (* ...and no wakeup is broadcast *)
```

A way to fix this is to temporarily prevent cancellation, but unfortunately we don't yet have scheduler agnostic and friendly mutexes and conditions nor a [scheduler agnostic way to prevent cancellation](https://github.com/polytypic/io#cancellation).

So, I'd recommend to add TODO notes to the code:
- TODO: Use scheduler friendly mutexes and condition variables
- TODO: Make operations cancellation safe

Pinging also @lyrm as this is a general future issue with lock based data structures.

lyrm:
Thanks for the work. 

Since there were issues with the mutex/condition modules from the Stdlib we began working on alternative using Picos (see PR #83). Consequently, I am closing this PR.

PR #79, opened 2023-07-12
added files for fine-grained linked list as a set

DeltaCube23:
Implementation of a Fine-grained linked list (as a set) based on optimistic synchronisation referred from section 9.6 in art of multiprocessor programming. The add function requires only 1 lock, and the remove function requires 2 locks. The linked list works like a set, with elements arranged in ascending order.

### DIFF
--- a/src/fine_list.ml
+++ b/src/fine_list.ml
@@ -1,0 +1,148 @@
+type 'a node = { value : 'a; mutable next : 'a node option; lock : Mutex.t }
+type 'a t = { head : 'a node }
+
+let create_node value nextptr =
+  { value; next = nextptr; lock = Mutex.create () }
+
+let create key = { head = create_node key None }
+
+(* find the previous node <= key *)
+let find_previous_add t key =
+  let rec aux prev next =
+    match next with
+    | Some node when node.value > key -> prev
+    | Some node -> aux node node.next
+    | None -> prev
+  in
+  aux t.head t.head.next
+
+(* find the previous node < key *)
+let find_previous_remove t key =
+  let rec aux prev next =
+    match next with
+    | Some node when node.value >= key -> prev
+    | Some node -> aux node node.next
+    | None -> prev
+  in
+  aux t.head t.head.next
+
+(* add new node in correct position *)
+let add t value =
+  (* x will point to the new node after insertion *)
+  let insert x =
+    if x.value = value && not (x == t.head) then (
+      Mutex.unlock x.lock;
+      false)
+    else
+      let new_node = create_node value x.next in
+      x.next <- Some new_node;
+      Mutex.unlock x.lock;
+      true
+  in
+  (* check if prev and cur are still in same position after locking *)
+  let rec validate prev cur =
+    Mutex.lock prev.lock;
+    let verify = find_previous_add t value in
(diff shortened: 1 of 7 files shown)


src/fine_list.ml
```diff
@@ -0,0 +1,148 @@
...
+    | Some node -> aux node node.next
+    | None -> prev
+  in
+  aux t.head t.head.next
+
+(* add new node in correct position *)
+let add t value =
+  (* x will point to the new node after insertion *)
+  let insert x =
+    if x.value = value && not (x == t.head) then (
```
polytypic, line 33: It is usually best to give the notion of equality over values as a parameter.

src/fine_list.ml
```diff
@@ -0,0 +1,148 @@
...
+let contains t value =
+  let rec validate prev cur =
+    Mutex.lock prev.lock;
+    let is_tail = ref false in
+    let to_find =
+      match cur with
+      | Some node -> node
+      | _ ->
+          is_tail := true;
+          create_node value None
```
polytypic, line 114: Hmm... Is the node created here used for something?

src/fine_list.ml
```diff
@@ -0,0 +1,148 @@
...
+    in
+    if !is_tail then (
```
polytypic, line 116: Could you avoid allocating a ref cell for the `is_tail` condition by inlining the continuation branches to the match:

```ocaml
match cur with
| None ->
    Mutex.unlock prev.lock;
    false
| Some to_find ->
    Mutex.lock to_find.lock;
    let verify = find_previous_remove t value in
    (* ... *)
```

src/fine_list.ml
```diff
@@ -0,0 +1,148 @@
...
+      Mutex.unlock prev.lock;
+      false)
+    else (
+      Mutex.lock to_find.lock;
+      let verify = find_previous_remove t value in
+      let temp = verify.next in
+      match temp with
+      | Some _ when temp == cur && prev == verify ->
+          let res = to_find.value = value in
```
polytypic, line 125: I believe the logic around here could be simplified and the equality check can be moved outside of the critical section.  Something like this:

```ocaml
      let verify = find_previous_remove t value in
      if prev == verify && cur == verify.next then begin
          Mutex.unlock prev.lock;
          Mutex.unlock to_find.lock;
          to_find.value = value (* better to use user defined equality *)
      end
      else begin
          Mutex.unlock prev.lock;
          Mutex.unlock to_find.lock;
          let again = find_previous_remove t value in
          validate again again.next
      end
```

The common prefix of both branches could also be factored out.

src/fine_list.ml
```diff
@@ -0,0 +1,148 @@
...
+let create_node value nextptr =
+  { value; next = nextptr; lock = Mutex.create () }
+
+let create key = { head = create_node key None }
+
+(* find the previous node <= key *)
+let find_previous_add t key =
+  let rec aux prev next =
+    match next with
+    | Some node when node.value > key -> prev
```
polytypic, line 13: Hmm... I wonder whether the logic and `add` and `remove` could be modified slightly so that only a single `find_previous` helper would be needed without impacting performance.

PR #80, opened 2023-07-12
added initial files for priority queue

DeltaCube23:
Implementation of a lockfree priority queue using a skiplist based on chapters 14 and 15 in art of multiprocessor programming and the following [paper](http://people.csail.mit.edu/shanir/publications/Priority_Queues.pdf).


lyrm:
I convert this to a draft PR as a lot of work is still due, in particular because the code of #65 is copied/pasted and adapted for a priority queue.

PR #81, opened 2023-07-13
Generate opam files automatically

Sudha247:


Sudha247:
Umm, as expected, byte-code only platforms (32-bit, ppc64) are timing out on the CI. I'll look into disabling some of the long running tests temporarily on those. I think we shouldn't mark the package itself as unavailable.

jmid:
IIUC, the remaining arm32 `failed to allocate domain` error is from https://github.com/ocaml-multicore/saturn/blob/main/test/ws_deque/test_ws_deque.ml where I suspect 16 `thieves` would need adjusting :thinking:

Sudha247:
Thank you for catching the `thieves`, @jmid! :smile:

lyrm:
Does this PR need a review ? Or what is missing to merge it ?

test/treiber_stack/dune
```diff
@@ -7,6 +7,11 @@
 (test
  (name treiber_stack_dscheck)
  (libraries atomic dscheck alcotest)
+ (enabled_if
+  (not
+   (and
+    (= %{arch_sixtyfour} false)
+    (= %{architecture} arm))))
```
lyrm, line 14: Why some datastructures need that, but not the `spsc` queue and the `mpsc` queue ?

lyrm:
Should we try to get the CI happy (!) ?

test/treiber_stack/dune
Sudha247: Some tests are quite long running and cause the CI to time out. I've only disabled those.

Sudha247:
I thought the CI was happy, but it suddenly seems to be acting out. We can ignore the bench failure as it was on `autumn`. The bench results on `fermat` appear here.

Sudha247:
All right, I suspect CI failures were due to some changes in ocaml-ci. It seems fine now. Can we merge this?

PR #83, opened 2023-07-14
A lock-free bounded queue

polytypic:
This implements a lock-free bounded queue &mdash; blocking using [domain-local-await](https://github.com/ocaml-multicore/domain-local-await/).

The data structure / algorithm is based on the Michael-Scott queue extended with length maintenance and caching of remaining capacity in such a way that additional contention is avoided in the happy paths.

TODO:
- [x] fix opam files to build on CI
- [x] use backoff from separate package
   - [x] fix dune-project
- [ ] `try_push`, `length` (can be done in `O(1)`), `is_empty`
- [ ] tests
- [ ] ability to change capacity
- [ ] benchmarking
- [ ] consider whether making wakeups fair is worth it (currently all awaiters are woken up so there is no fairness)
- [ ] documentation

### DIFF
--- a/src_lockfree/cue.ml
+++ b/src_lockfree/cue.ml
@@ -1,0 +1,260 @@
+(* Copyright (c) 2023-2024, Vesa Karvonen <vesa.a.j.k@gmail.com>
+
+   Permission to use, copy, modify, and/or distribute this software for any
+   purpose with or without fee is hereby granted, provided that the above
+   copyright notice and this permission notice appear in all copies.
+
+   THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES WITH
+   REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+   AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+   INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+   LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+   OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+   PERFORMANCE OF THIS SOFTWARE. *)
+
+open Picos
+module Atomic = Multicore_magic.Transparent_atomic
+
+type ('a, _) node =
+  | Null : ('a, [> `Null ]) node
+  | Node : {
+      mutable _next : 'a link;
+      mutable value : 'a;
+      mutable capacity : int;
+      mutable counter : int;
+    }
+      -> ('a, [> `Node ]) node
+
+and 'a link = Link : ('a, [< `Null | `Node ]) node -> 'a link [@@unboxed]
+
+external link_as_node : 'a link -> ('a, [ `Node ]) node = "%identity"
+
+external next_as_atomic : ('a, [< `Node ]) node -> 'a link Atomic.t
+  = "%identity"
+
+let[@inline] get_capacity (Node r : (_, [< `Node ]) node) = r.capacity
+
+let[@inline] set_capacity (Node r : (_, [< `Node ]) node) value =
+  r.capacity <- value
+
+let[@inline] get_counter (Node r : (_, [< `Node ]) node) = r.counter
+
+let[@inline] set_counter (Node r : (_, [< `Node ]) node) value =
+  r.counter <- value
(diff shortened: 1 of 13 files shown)


src/cue.ml
```diff
@@ -0,0 +1,244 @@
...
+
+(* *)
+
+let rec peek_opt t =
+  let head = Atomic.get t.head in
+  match fenceless_get_next head with
+  | Null -> None
+  | Node r as node ->
+      let value = r.value in
+      if Atomic.get t.head != node then peek_opt t else Some value
```
lyrm, line 124: ` if Atomic.get t.head != node then ...` 

There is a small bug here, it should be : 

 `if Atomic.get t.head != head then ...`

Same in `peek`.

lyrm: There is still some bugs between `peek_opt` and `pop_opt`.

                                  |                        
                        Try_push 42 : true               
                                  |                        
               .------------------------------------.
               |                                    |                        
      Peek_opt : Some (0)            Pop_opt : Some (42) 

I will continue investigating !

polytypic: Good catch!  Yes, this is a bug.  What would ideally be needed here is a(n acquire) fence to prevent the `r.value` read from being reordered after the `Atomic.get`, but OCaml currently doesn't provide a direct way to do _only_ that.   Some alternatives:
- `value` could be turned into an atomic and then `r.value` would be replaced with `Atomic.get r.value`.   This would be quite expensive.
- Instead of `Atomic.get t.head == head` one could use `Atomic.compare_and_set t.head head head`.  This is also somewhat expensive, because means that the `t.head` location is actually written to.
- In practise, but not in the OCaml memory model, which doesn't match actual hardware, it is also possible to generate fences by accessing an unrelated location.  For example, one could define
  ```ocaml
  let fence () = Atomic.set (Atomic.make 0) 0
  ```
  and use that to write
  ```ocaml
  let value = r.value in
  fence ();
  if Atomic.get t.head != head then ...
  ```

lyrm: The `fence` trick seems to work !

polytypic: This may actually be a compiler bug.

The OCaml 5 compiler [definitely has a bug](https://github.com/ocaml/ocaml/issues/12713) related to an incorrect optimization of `Atomic.get` operations.

What remains to consider is whether it is actually allowed to reorder a nonatomic load that precedes an atomic load to happen after the atomic load.  I suspect it is not actually allowed.  See the explanation of fences in this [post](https://preshing.com/20130922/acquire-and-release-fences/) about fences in C++11:

> An acquire fence prevents the memory reordering of any read which precedes it in program order with any read or write which follows it in program order.

This, of course, is not exactly the same as the semantics of `Atomic.get` in OCaml 5, but according to the above explanation the original code

```ocaml
let value = r.value in
(* The above read should not be allowed to happen after
    the read below as the read below has an acquire fence. *)
if Atomic.get t.head != head then ...
```

would be correct.  I'll try to redecipher the [Bounding Data Races in Space and Time](https://kcsrk.info/papers/pldi18-memory.pdf) paper to find definite word on this.

polytypic: Assuming this is just a compiler bug, then the following workaround would be sufficient:

```ocaml
let value = r.value in
(* The [Sys.opaque_identity] below prevents OCaml 5 from
    optimizing the repeated load away. *)
if Atomic.get (Sys.opaque_identity t.head) != head then ...
```

polytypic:
FYI, I will squash this PR to a single commit to make this PR a little bit easier to work with.

src/cue.ml
polytypic: Just to summarize, before I mark this as resolved, this turned out to be a compiler bug, which has been fixed in OCaml 5.2.

polytypic:
FYI, I changed this to use Picos instead of DLA.

src_lockfree/cue.ml
```diff
@@ -0,0 +1,260 @@
...
+          match Trigger.await trigger with
+          | None -> ()
+          | Some (exn, bt) ->
+              signal_all t.tail_waiters;
+              Printexc.raise_with_backtrace exn bt
+      end;
+      peek t
+  | Link (Node r) ->
+      let value = r.value in
+      if Atomic.get t.head != old_head then peek t else value
```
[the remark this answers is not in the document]
art-w: If there's a parallel `pop` in progress, the `r.value` will be set to `Obj.magic ()` (line 169). So without the conditional we don't know if the `r.value` we read was correct (or a null) when the head has changed (?)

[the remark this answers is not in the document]
lyrm: Yes, of course, thanks !

lyrm:
#160 has been merged and adds this implementation without the bounding part. For now, we are not planning to add a dependency to picos so I close this PR. 

Thanks @polytypic  for the work !

PR #89, opened 2023-07-27
Add .nojekyll

lyrm:
Add `.nojekyll` file.


polytypic (approved):
🚀

Sudha247:
I may be unfamiliar with this, but why do we need to bypass jekyll processing? Is it recommended for odoc generated files?

lyrm:
It's a recommendation from @polytypic . Right now the documentation does not look good because (from what I understand) `jekyll` has issues with file names with `_` in it.

It is also how it is fixed in [eio](https://github.com/ocaml-multicore/eio/blob/gh-pages/.nojekyll).

Sudha247:
Sounds good to me!

polytypic:
Just noting that the CI failures should be unrelated to the addition of the `.nojekyll` file.

Also, I've added this `.nojekyll` thing to many repos already.

PR #90, opened 2023-07-30
Lazy skiplist

koonwen:
This PR adds a lock-based lazy-skiplist discussed [here](https://github.com/ocaml-multicore/saturn/issues/87)

More tests and benchmarks to come

### DIFF
--- a/src_lockbased/lazy_skiplist.ml
+++ b/src_lockbased/lazy_skiplist.ml
@@ -174,8 +174,15 @@ module Node (V : Compare) = struct
     | Node { lock; _ } -> RMutex.unlock lock
 end
 
+(* Caveats: If the hash algorithm encounters a collision, the element
+   will fail to be added to the list *)
 module Make (V : Compare) = struct
   module Node = Node (V)
+
+  type t = {
+      head : Node.t;
+      tail : Node.t;
+    }
 
   let maxlevel = 32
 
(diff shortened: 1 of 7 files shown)


src_lockbased/lazy_skiplist.ml
```diff
@@ -0,0 +1,360 @@
...
+      primitive = Mutex.create ();
+      wait      = Condition.create ();
+      ownership = None
+    }
+
+  (**
+     Attempt to acquire the mutex, waiting indefinitely
+   *)
+  let lock m =
+    let id = Thread.id (Thread.self ()) in
```
polytypic, line 49: Just noting that this will not work, in general, if you have a scheduler that allows fibers to move from one (domain and) (sys)thread to another (like Domainslib) or allows multiple fibers to run on a single thread (like Eio).  Unfortunately we don't yet have a scheduler agnostic way to get the id of a fiber.

koonwen: Thanks for the review! @polytypic 

I must confess that I find the many names & levels of threads like `systhreads, pthreads, domains, lightweight threads, fibers, etc...` quite confusing. So just to say the obvious thing to confirm that I've understood you correctly:

`Thread.id (Thread.self ())` doesn't work as we would expect under effect-based schedulers because they use fibers which are simulated lightweight threads by the runtime and are different from the pthreads used internally by the `Thread` module. Therefore, if for example we use Eio scheduler, all fibers that call `Thread.id` returns the same one value which is the id of the "main" thread? In the case of using Domainslib which moves fibers between domains, the value of `Thread.id` just depends on which domain it is called from regardless of the work fiber is doing

polytypic: Yes, I think you explained it well. 👍

Also, I haven't really reviewed this PR, yet.  Just a random observation from taking a peek at the code.

lyrm:
Thanks for the work ! 

I am closing this PR as there is no more reasons to add  this lock-based skip list, that is outperformed by the lock-free one and has all the issues of the stdlib mutexes.

PR #92, opened 2023-09-07
`pop_opt`, `peek`, `peek_opt`  functions for `queue`

lyrm:
This PR renames and adds functions to the single consumer single producer queue, the michael-scott queue, the treiber stack and the multiple producer single consumer queue. 
- `pop` is renamed `pop_opt` 
- the new `pop`  raises `Empty` if the queue is empty
- `peek` and `peek_opt` have been added as well (except for the stack)

Other notes :
- About work stealing deque : I did not add the `peek` functions, as it seems useless (it is useful mostly with a single consumer). 
- About michael-scott queue : I removed `clean_until` function as it was no atomic. 

TODO : 
- [x] ~Rename `pop` to `pop_opt` for relaxed-queue~,    
- [x] Rename `pop` to `pop_opt` for the work-stealing deque
- [x] Answer the question :  should `pop` and `pop_opt` be almost copy-paste function or should we use the  `inline` keyword (see comment in code)

### DIFF
--- a/src_lockfree/michael_scott_queue.ml
+++ b/src_lockfree/michael_scott_queue.ml
@@ -31,7 +31,9 @@ let create () =
 
 let is_empty { head; _ } = Atomic.get (Atomic.get head) == Nil
 
-let pop { head; _ } =
+exception Empty
+
+let pop_opt { head; _ } =
   let b = Backoff.create () in
   let rec loop () =
     let old_head = Atomic.get head in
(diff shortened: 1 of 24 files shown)


src_lockfree/michael_scott_queue.ml
```diff
@@ -45,6 +47,27 @@ let pop { head; _ } =
...
+  let rec loop () =
+    let old_head = Atomic.get head in
+    match Atomic.get old_head with
+    | Nil -> raise Empty
+    | Next (value, next) when Atomic.compare_and_set head old_head next -> value
+    | _ ->
+        Backoff.once b;
+        loop ()
+  in
+  loop ()
```
lyrm, line 61: This is the same function as `pop_opt` except for very small changes. Should I use `[@inline]` and a generic pop function ?

src_lockfree/spsc_queue.ml
```diff
@@ -48,7 +48,27 @@ let push { array; head; tail; mask; _ } element =
     Array.set array (tail_val land mask) (Some element);
     Atomic.set tail (tail_val + 1))
 
+let[@inline] try_push t element =
+  try
+    push t element;
+    true
+  with Full -> false
```
lyrm, line 55: Here I try the `[@inline]`. But I don't think this is the right way to use it. Any help will be welcomed.

lyrm (commented):
Comments about how and where to use `[@inline]`.

Sudha247 (commented):
LGTM. Minor typo below.

PR #94, opened 2023-09-22
Remove 'name' field from benchmark results

Sudha247:
@punchagan pointed out that the current-bench interface is adding another level of abstraction as the top-level json has a name field. Revoming this would get the results to show up on the front-page of the current-bench run.


Sudha247:
Looks like this is showing the graphs.

lyrm (approved):
This looks good. Thanks !

PR #106, opened 2023-11-19
Execute benchmarks in release mode

polytypic:
It doesn't typically make a big difference in OCaml to run things in release mode, but let's still run benchmarks in release mode and avoid potential issues later where the benchmark results might be misleading due to not running in release mode.


Sudha247 (approved):
Thanks.

PR #107, opened 2023-11-21
Replacing local backoff module by backoff library.

lyrm:
This PR replaces the internal `backoff` module by the [backoff](https://github.com/ocaml-multicore/backoff/pulls) library (whose code originated from `saturn` and `kcas`). 

`backoff` benches have been removed from Saturn and are now in the `backoff` library.


src_lockfree/michael_scott_queue.ml
```diff
@@ -35,30 +35,30 @@ exception Empty
 
 let pop_opt { head; _ } =
   let b = Backoff.create () in
```
polytypic, line 37: Hmm... Probably won't make a difference (assuming the compiler is able to constant fold the `Backoff.create ()` call), but it might be a bit safer / more future proof to use the constant `Backoff.default`.  (The same for all the other uses of `Backoff.create ()`.)

src_lockfree/treiber_stack.ml
```diff
@@ -14,17 +14,16 @@ let push q v =
   let new_node = Next (v, head) in
   if Atomic.compare_and_set q.head head new_node then ()
   else
-    let b = Backoff.create () in
-    Backoff.once b;
+    let b = Backoff.create () |> Backoff.once in
```
polytypic, line 17: Hmm... I'd recommend considering to rewrite the code here a bit more.  A kind of (recursive) worker- (non-recursive) wrapper idiom:

```ocaml
let rec push backoff q v =
  ...

let push q v = push Backoff.default q v
```

I believe this would make the code a bit shorter / simpler and should perform reasonably well.

src_lockfree/michael_scott_queue.ml
```diff
@@ -34,31 +34,31 @@ let is_empty { head; _ } = Atomic.get (Atomic.get head) == Nil
...
     let old_head = Atomic.get head in
     match Atomic.get old_head with
     | Nil -> None
     | Next (value, next) when Atomic.compare_and_set head old_head next ->
         Some value
     | _ ->
-        Backoff.once b;
-        loop ()
+        let b = Backoff.once b in
+        loop b
```
polytypic, line 46: I'd probably shorten this to `loop (Backoff.once b)`, but this is mostly stylistic.

src_lockfree/michael_scott_queue.ml
```diff
@@ -34,31 +34,31 @@ let is_empty { head; _ } = Atomic.get (Atomic.get head) == Nil
...
   in
-  loop ()
+  loop b
```
polytypic, line 48: I'd probably drop the `let b = Backoff.default` and write `loop Backoff.default` here, but this is mostly stylistic.

polytypic (approved):

PR #111, opened 2023-12-11
Updated the MPSC queue implementation with the optimized version from picos.

expl-rel:
Incorporated several functions to align with the Saturn implementation. 
Still in progress: Adjustments needed as tests are currently failing; need to check and correct accordingly.


dune-project
```diff
@@ -27,7 +28,9 @@
   (domain_shims (>= 0.1.0))
   (alcotest (and (>= 1.7.0) :with-test))
   (qcheck (and (>= 0.18.1) :with-test))
+  (backoff (>= 0.1.0))
```
polytypic, line 31: If you run `dune build` (for example) then `dune` will regenerate the `saturn_lockfree.opam` file and this dependency should then be included in the opam file.

src_lockfree/mpsc_queue.mli
```diff
@@ -1,69 +1,67 @@
...
+    
+        @raise Closed if [q] is closed and empty. *)
+    
+    val push_head : 'a t -> 'a -> unit
+    (** [push_head q v] adds the element [v] at the head of the queue
+        [q]. This can only be used by the consumer (if run in parallel
+        with {!pop}, the item might be skipped).
+    
+        @raise Closed if [q] is closed and empty. *)
+
```
polytypic, line 67: This indentation in this does not seem to match what [`ocamlformat`](https://github.com/ocaml-ppx/ocamlformat) would produce.  Install `ocamlformat` if you haven't already by running `opam install ocamlformat` (we're using the latest version).  You can probably configure your editor to use `ocamlformat` to format files on save.  You can also run `dune fmt` to format files in the project.

src_lockfree/mpsc_queue.mli
```diff
@@ -1,69 +1,67 @@
...
+    (** [create ()] returns a new empty queue. *)
+    
+    val is_empty : 'a t -> bool
+    (** [is_empty q] is [true] if calling [pop] would return [None].
+    
+        @raise Closed if [q] is closed and empty. *)
+    
+    val close : 'a t -> unit
+    (** [close q] marks [q] as closed, preventing any further items from
+        being pushed by the producers (i.e. with {!push}).
```
polytypic, line 19: Note that with the new implementation `close` can only be safely called by the consumer.  Basically any operation whose internal implementation touches the `head` can only be safely called by the consumer.  It would be useful to e.g. group the operations in this signature file into two sections where one section contains operations that may only be called by the consumer and the other section contains operations that can also be called by the producers.

polytypic: Note that, for example, the comment at https://github.com/ocaml-multicore/saturn/blob/e3902f51a0fd8fa950c1a99502955714d76eae76/test/mpsc_queue/qcheck_mpsc_queue.ml#L3-L13 no longer holds and should be updated.

src_lockfree/mpsc_queue.ml
```diff
@@ -44,10 +44,10 @@ let rec pop t =
   | x :: xs -> t.head <- xs; x
   | Closed -> raise Closed
   | Open ->
-    (* We know the tail is open because we just saw the head was open
-       and we don't run concurrently with [close]. *)
+    (* We know the tail is open because we just saw the head was open *)
+    (* we can run concurrently with close *)
     match Atomic.exchange t.tail Open with
-    | Closed -> assert false
+    | Closed -> pop t
```
polytypic, line 50: Hmm... This shouldn't happen when the queue is used correctly.  Note that `close` can no longer be called by producers and the existing tests might be assume that it is safe to call `close` from producers.

Instead of an `assert false` it might be useful to use `failwith` here with a short error message that explains that this shouldn't happen unless `close` was called from a producer.

PR #115, opened 2023-12-17
Prefer `Backoff` over `Domain.cpu_relax` and `domain_shims`

polytypic:
This PR replaces a use of `Domain.cpu_relax ()` with `Backoff` and implements a `Domain` module with `cpu_relax = Thread.yield` on OCaml 4.  This allows dropping the dependency to `domain_shims`.  I would recommend not depending on `domain_shims` in cases where the dependency is easy to avoid (except on internal tests, for example).

`domain_shims` currently (implicitly) depends on `threads`, for example, which, I believe, means that it does not directly build with `js_of_ocaml`.  Some of the detailed decisions in `domain_shims` also may not be exactly right for all use cases.  For example, if you want to use `Domain.DLS` to avoid contention, then the implementation in `domain_shims` is not necessarily what you want, because it is fundamentally a `TLS` (thread-local storage) rather than a `DLS` (domain-local storage).

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -12,7 +12,7 @@
  (synopsis "Collection of parallelism-safe data structures for Multicore OCaml")
  (depends
   (ocaml (>= 4.12))
-  (domain_shims (>= 0.1.0))
+  (domain_shims (and (>= 0.1.0) :with-test))
   (saturn_lockfree (= :version))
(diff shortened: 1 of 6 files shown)


Sudha247 (approved):
LGTM!

PR #117, opened 2023-12-20
Lock free hash table

lyrm:
Following the algorithm implemented in PR #62, this PR implements a hash table with better types and numerous optimizations to the algorithm. It aims to provide a full resizable hash table (both shrinkable and growable) and an API as close as possible to the stdlib hash table. 

## Algorithm in brief
The general idea of the algorithm is the same than the hash set in the previous PR (see also the chapter 13.3 of the Art of Multicore Programming, 2nd edition) : a single sorted linked list  contains all the keys. The hash table itself is an array of buckets that simply serve as  shortcuts (i.e. pointers)  to some special nodes in the sorted linked list. As it is sorted, a `find` call only needs to pass though the nodes in between two of these special nodes to determine if a key is in the hash table, meaning if the hash table is grown properly, `find` has a O(1) complexity, as expected for an hash table. 

The key advantage of the algorithm is that growing or shrinking it does not require to move nodes from one bucket to another : it just requires to add (or remove) some "special" shortcuts nodes. All the nodes actually containing data are sorted in a way they don't need to be moved and stay untouched during a grow/shrink operation.

## Implemented changes to the algorithm
 All main changes to the algorithm are/will be mentioned (and sometime even explained !) in this [gist](https://gist.github.com/lyrm/030091bb8ec17f36f9808c886118d2db).  

## Current status
- sorted linked list with multiple bindings
   - implemented operations (same semantics as the stdlib hash table) : `add`, `remove`, `replace`, `find_all`, `find_opt`/`find`   
   - tests with `stm` for these functions
   - tests with `dscheck` for `remove` and `add`
- empty node are managed as expected (to replace the dummy node of the original algorithm) : they can be added and removed (with `add_empty` and `try_remove ~empty=false`) and have no impact on the size.  

## Work in progress
The linked list have most of the needed functionalities. Next step : hash table !


lyrm:
@polytypic  I pushed my solution to make the counter work. If you could  have a look, that would be great :)

test/hmap/llist_dscheck.ml
```diff
@@ -0,0 +1,242 @@
+let try_add ll elt = Llist.try_add ll elt ()
+
+module Sint = Set.Make (struct
+  type t = int
+
+  let compare = compare
```
polytypic, line 4: `Set.Make (Int)` would be preferable.  The `compare` here refers to the polymorphic compare, which is much slower than `Int.compare`.

test/hmap/llist_dscheck.ml
```diff
@@ -0,0 +1,242 @@
...
+end)
+
+let xor a b = match (a, b) with true, false | false, true -> true | _ -> false
```
polytypic, line 13: You could also use `let xor a b = a != b`.

PR #118, opened 2023-12-21
Updated the MPSC queue implementation with the version from picos and the tests

expl-rel:
The queue is implemented based on the version from [Picos](https://github.com/ocaml-multicore/picos/blob/main/test/lib/foundation/mpsc_queue.ml) and adding certain other functions.
The tests are changed (removed) because some of the functions like `close`,` is_empty`, `peek` are made only consumer operations in the case of a MPSC queue.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,6 +12,7 @@ on:
 jobs:
   windows:
     strategy:
+      fail-fast: false
       matrix:
         ocaml-compiler:
           - ocaml.5.0.0,ocaml-option-mingw
@@ -48,6 +49,7 @@ jobs:
 
   build-4_x:
     strategy:
+      fail-fast: false
       matrix:
         os:
           - ubuntu-latest


src_lockfree/mpsc_queue.ml
```diff
@@ -1,152 +1,145 @@
...
+  match Atomic.get t.tail with
+  | T Tclosed -> raise Closed
+  | before ->
+      let after = T (Snoc (before, x)) in
+      if not (Atomic.compare_and_set t.tail before after) then push t x
+
+(*  *)
+let rec rev_to (head : (_, [< `Cons ]) head) = function
+  | T Topen -> head
+  | T Tclosed -> assert false
```
polytypic, line 39: I can see now that this is probably what you mentioned as a case that could still be eliminated.  Yes, this match case `T Tclosed` and the other one in `rev` below can also be eliminated (after changing the type appropriately). 👍

src_lockfree/mpsc_queue.ml
```diff
@@ -1,152 +1,145 @@
...
+  Multicore_magic.copy_as_padded { tail; head }
+
+(* let push t x = t.tail <- T (Snoc (t.tail, x)) *)
+
+let rec push t x =
+  match Atomic.get t.tail with
+  | T Tclosed -> raise Closed
+  | before ->
+      let after = T (Snoc (before, x)) in
+      if not (Atomic.compare_and_set t.tail before after) then push t x
```
polytypic, line 34: It would be good to have backoff at this point in case the `compare_and_set` fails (i.e. returns false).

Note that Saturn recently switched to using the [backoff](https://github.com/ocaml-multicore/backoff) library.  If you rebase the branch against `main` you will then get access to the `Backoff` module from the backoff library and can implement backoff to deal with contention on push.

src_lockfree/mpsc_queue.ml
```diff
@@ -1,152 +1,145 @@
...
+      match Atomic.exchange t.tail (T Topen) with
+      | T Topen -> None
+      | T Tclosed ->
+          failwith
+            "This cannot happen, maybe you are running close concurrently with \
+             [pop]"
+      | T (Snoc _ as tail) -> (
+          match rev_pop tail with
+          | Cons (x, xs) ->
+              t.head <- xs;
```
polytypic, line 75: Here (and in `pop`) performance could be improved slightly by skipping the write `t.head <- xs` in case `xs == H Hopen`, e.g.

```diff
          match rev_pop tail with
+         | Cons (x, H Hopen) -> Some x
          | Cons (x, xs) ->
              t.head <- xs;
```

Writing to a potentially pointer containing field in OCaml is relatively expensive as it requires a write barrier.

src_lockfree/mpsc_queue.ml
```diff
@@ -1,152 +1,145 @@
...
+  | H (Cons (x, _)) -> x
+  | H Hclosed -> raise Closed
+  | H Hopen -> (
+      match Atomic.get t.tail with
+      | T Topen -> raise Empty
+      | T Tclosed ->
+          failwith
+            "This cannot happen, maybe you are running close concurrently with \
+             [close]"
+      | T (Snoc _ as tail) -> ( match rev_pop tail with (Cons (x, _)) -> x))
```
polytypic, line 115: This can be a potential performance issue, because repeated calls to `peek` keep on performing the `rev_pop` call.  IOW, it would make sense to move the elements from the tail to the head in peek so that peek will not repeatedly reverse the tail.

PR #120, opened 2024-01-12
Rewrite or refactor benchmarks using multicore-bench

polytypic:
This PR changes the benchmarks to use the [multicore-bench](https://github.com/ocaml-multicore/multicore-bench) framework.

This PR also adds a `bench.Dockerfile` to run the benchmarks on OCaml 5.1 (rather than [the default OCaml 5.0](https://github.com/ocurrent/current-bench/blame/9825e36d34999adf79e9a57882b19cf3d51f3785/doc/user_manual.md#L21-L25)).

See also:
* Similar [PR in Kcas](https://github.com/ocaml-multicore/kcas/pull/179).

---

It is clear that many of the data structure implementations currently suffer from false sharing and some other performance pitfalls.  I briefly experimented with some optimizations (avoiding false sharing, avoiding indirections, avoiding contention by using a caching mechanism) and was able to e.g. get the spsc queue to perform more than order of magnitude better (from 15M/s to over 200M/s) on my M3 Max laptop.  I'm, however, intentionally leaving the proper implementation of these optimizations for future work.


saturn.opam.template
```diff
@@ -0,0 +1,3 @@
+pin-depends: [
+  [ "multicore-bench.dev" "git+https://github.com/ocaml-multicore/multicore-bench#bdbdc6787d670725349c880581be770638fbb05b" ]
```
polytypic, line 2: The multicore-bench package needs to be published and this pin-depend removed before merging.

lyrm (approved):
Thanks for these PR. `Multicore-bench` is a way more readable framework than what we had before. That will make benchmarks more useful and hopefully less painful to write :) 

Also, I like the `ws_queue` bench with Fibonacci. I will probably use this to write an example on how to use this queue.

PR #121, opened 2024-01-19
WIP: Optimize SPSC queue

polytypic:


lyrm (commented):
I am unsure the commit _Avoid float array pessimization_ and a bit of the previous one (removing the indirection of the array and thus doing initialization with `Obj.magic`) are a good idea. They introduce a lot of `Obj.magic` which I feel we should be careful with, especially because it could cause issues with flambda. 

Also, what happen if you are actually putting float in the queue ? I am guessing the array float special case will not be triggered ? 

I have no immediate solution for  float array pessimization, but for initialization, we could have a `make : size_exponent:int -> 'a ->  'a` function instead of the `create` function that requires the user to provide an inhabitant  for type `'a`.  

It may just move the issue to the user lever that may want to either use the same trick with `Obj.magic` or use and `option` type  though...

lyrm:
Thanks for the PR ! All the optimizations (except maybe the ones mentioned in my previous message :p) are great and thanks again for making the git history so readable. And obviously, the gain in performance is amazing !

lyrm:
I should have added that the optimization with the cashes is very cool !!

lyrm:
The changes proposed in this PR have already been merged in two different PRs (#133  and #135) that separate "safe" and "unsafe" optimizations.  Thanks @polytypic  for all this work!

PR #122, opened 2024-01-21
Optimize Michael-Scott queue and remove `snapshot` mechanism

polytypic:
This PR optimizes the Michael-Scott queue implementation.  The main optimizations are avoiding false sharing and then optimization of the node representation, which effectively halves the length of the linked list of nodes by "inlining" the atomic into the node record and improves performance dramatically.

This PR also removes the snapshot mechanism.  The snapshot mechanism is not used in any tests in this repository.  Furthermore, the snapshot mechanism cannot be made to work (efficiently) as such with the traditional Michael-Scott queue representation where the head points to a node whose value has been taken from the queue.  That is because, to avoid space leaks, the value in the head node must be overwritten and that simply breaks the snapshot mechanism.  The previous implementation avoided this problem by using an inefficient node representation that effectively made the linked list of nodes twice as long and operations on the queue twice as slow.

Here is a run of the benchmarks on my M3 Max before the optimizations:
```json
➜  saturn git:(rewrite-bench) ✗ dune exec --release -- ./bench/main.exe -budget 1 'free Queue' | jq '[.results.[].metrics.[] | select(.name | test("over")) | {name, value}]'
[                                    
  {
    "name": "messages over time/one domain",
    "value": 25.832570841950652
  },
  {
    "name": "messages over time/1 nb adder, 1 nb taker",
    "value": 30.775546858386015
  },
  {
    "name": "messages over time/1 nb adder, 2 nb takers",
    "value": 28.61298006857959
  },
  {
    "name": "messages over time/2 nb adders, 1 nb taker",
    "value": 26.324158773037244
  },
  {
    "name": "messages over time/2 nb adders, 2 nb takers",
    "value": 26.72249173339718
  }
]
```

And here after the optimizations:
```json
➜  saturn git:(optimize-michael-scott-queue) ✗ dune exec --release -- ./bench/main.exe -budget 1 'free Queue' | jq '[.results.[].metrics.[] | select(.name | test("over")) | {name, value}]'
[                                    
  {
    "name": "messages over time/one domain",
    "value": 49.09980418998089
  },
  {
    "name": "messages over time/1 nb adder, 1 nb taker",
    "value": 98.92424842899835
  },
  {
    "name": "messages over time/1 nb adder, 2 nb takers",
    "value": 73.48287261205158
  },
  {
    "name": "messages over time/2 nb adders, 1 nb taker",
    "value": 79.05834136653924
  },
  {
    "name": "messages over time/2 nb adders, 2 nb takers",
    "value": 70.05359476527885
  }
]
```

Here is a run of the benchmarks without the inlined node representation:
```json
➜  saturn git:(optimize-michael-scott-queue) ✗ dune exec --release -- ./bench/main.exe -budget 1 'free Queue' | jq '[.results.[].metrics.[] | select(.name | test("over")) | {name, value}]'
[                                    
  {
    "name": "messages over time/one domain",
    "value": 27.799662289702503
  },
  {
    "name": "messages over time/1 nb adder, 1 nb taker",
    "value": 46.095344138917085
  },
  {
    "name": "messages over time/1 nb adder, 2 nb takers",
    "value": 42.19755884308922
  },
  {
    "name": "messages over time/2 nb adders, 1 nb taker",
    "value": 63.98353506270374
  },
  {
    "name": "messages over time/2 nb adders, 2 nb takers",
    "value": 58.21147004626938
  }
]
```

As you can see, on the `one domain` and `1 nb adder, 1 nb taker` configurations the inlined node representation is roughly twice as fast!  On the more contentious configurations, the additional cores likely both help to overcome some of the additional overhead (when not using the inlined representation) and the additional contention brings down the performance of inlined representation so the performance difference is not quite as dramatic.

You can run `cp test/michael_scott_queue/michael_scott_queue_node.ml src_lockfree` or `git reset --hard HEAD~1` to use the non-inlined node representation:

```diff
modified   src_lockfree/michael_scott_queue_node.ml
@@ -3,12 +3,10 @@ module Atomic = Transparent_atomic
 type ('a, _) t =
   | Nil : ('a, [> `Nil ]) t
   | Next : {
-      mutable next : ('a, [ `Nil | `Next ]) t;
+      next : ('a, [ `Nil | `Next ]) t Atomic.t;
       mutable value : 'a;
     }
       -> ('a, [> `Next ]) t
 
-let[@inline] make value = Next { next = Nil; value }
-
-external as_atomic : ('a, [ `Next ]) t -> ('a, [ `Nil | `Next ]) t Atomic.t
-  = "%identity"
+let[@inline] make value = Next { next = Atomic.make Nil; value }
+let[@inline] as_atomic (Next r : ('a, [ `Next ]) t) = r.next
```

### DIFF
--- a/src_lockfree/michael_scott_queue.ml
+++ b/src_lockfree/michael_scott_queue.ml
@@ -16,82 +16,92 @@
  * OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
  *)
 
-(* Michael-Scott queue *)
-
-type 'a node = Nil | Next of 'a * 'a node Atomic.t
+module Node = Michael_scott_queue_node
+module Atomic = Node.Atomic
 
 type 'a t = {
-  head : 'a node Atomic.t Atomic.t;
-  tail : 'a node Atomic.t Atomic.t;
+  head : ('a, [ `Next ]) Node.t Atomic.t;
+  tail : ('a, [ `Next ]) Node.t Atomic.t;
 }
 
 let create () =
-  let next = Atomic.make Nil in
-  let head = Atomic.make next |> Multicore_magic.copy_as_padded in
-  let tail = Atomic.make next |> Multicore_magic.copy_as_padded in
+  let node = Node.make (Obj.magic ()) in
+  let head = Atomic.make node |> Multicore_magic.copy_as_padded in
+  let tail = Atomic.make node |> Multicore_magic.copy_as_padded in
   { head; tail } |> Multicore_magic.copy_as_padded
 
-let is_empty { head; _ } = Atomic.get (Atomic.get head) == Nil
+let is_empty t = Atomic.get (Node.as_atomic (Atomic.get t.head)) == Nil
 
 exception Empty
 
-let pop_opt { head; _ } =
-  let b = Backoff.default in
-  let rec loop b =
-    let old_head = Atomic.get head in
-    match Atomic.get old_head with
-    | Nil -> None
-    | Next (value, next) when Atomic.compare_and_set head old_head next ->
-        Some value
-    | _ ->
-        let b = Backoff.once b in
-        loop b
-  in
-  loop b
+type ('a, _) poly = Option : ('a, 'a option) poly | Value : ('a, 'a) poly
 
-let pop { head; _ } =
-  let b = Backoff.default in
-  let rec loop b =
-    let old_head = Atomic.get head in
-    match Atomic.get old_head with
-    | Nil -> raise Empty
-    | Next (value, next) when Atomic.compare_and_set head old_head next -> value
-    | _ ->
-        let b = Backoff.once b in
-        loop b
-  in
-  loop b
-
-let peek_opt { head; _ } =
+let rec pop_as :
+    type a r. (a, [ `Next ]) Node.t Atomic.t -> Backoff.t -> (a, r) poly -> r =
+ fun head backoff poly ->
   let old_head = Atomic.get head in
-  match Atomic.get old_head with Nil -> None | Next (value, _) -> Some value
+  match Atomic.get (Node.as_atomic old_head) with
+  | Nil -> begin match poly with Value -> raise Empty | Option -> None end
+  | Next r as new_head ->
+      if Atomic.compare_and_set head old_head new_head then begin
+        match poly with
+        | Value ->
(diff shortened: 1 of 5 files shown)


src_lockfree/michael_scott_queue_node.ml
```diff
@@ -3,10 +3,12 @@ module Atomic = Transparent_atomic
...
+      mutable next : ('a, [ `Nil | `Next ]) t;
       mutable value : 'a;
     }
       -> ('a, [> `Next ]) t
 
-let[@inline] make value = Next { next = Atomic.make Nil; value }
-let[@inline] as_atomic (Next r : ('a, [ `Next ]) t) = r.next
+let[@inline] make value = Next { next = Nil; value }
+
+external as_atomic : ('a, [ `Next ]) t -> ('a, [ `Nil | `Next ]) t Atomic.t
```
lyrm, line 13: `next_as_atomic` seems like a better name for readability (and I think it is the name you usually use). Also, I won't be again a comment in the code about this line :)

lyrm (commented):
Thanks for the PR ! 
I made a few minor comments (including one related to the use of `Obj.magic` that seems safer here then in the `Spsc_queue` optimization PR). 

I really like the refactoring work to factorize the exception/option version of a function !

src_lockfree/michael_scott_queue.ml
```diff
@@ -16,80 +16,92 @@
...
-
-type 'a cursor = 'a node
-
-let snapshot { head; _ } = Atomic.get (Atomic.get head)
-let next = function Nil -> None | Next (a, n) -> Some (a, Atomic.get n)
+  let link = Node.as_atomic old_tail in
+  if Atomic.compare_and_set link Nil new_node then begin
+    if not (Atomic.compare_and_set tail old_tail new_node) then
+      let backoff = Backoff.once Backoff.default in
+      fix_tail tail new_node backoff
```
polytypic, line 103: Hmm... I think that calling `fix_tail` at this point is atually unnecessary.  It should be safe to simplify the above to:

```ocaml
Atomic.compare_and_set tail old_tail new_node |> ignore
```

The only reason for the CAS to fail _at this point_ should be that some other thread already pushed another value  into the queue and this means we don't need to worry about updating the tail anymore.

lyrm:
Thanks @polytypic  for this work. It has been merged via PR #146 .

PR #128, opened 2024-02-20
Upgrade DSCheck to allow testing on OCaml 4 with OCaml-CI

polytypic:
This PR upgrades to use [new version of the DSCheck package](https://github.com/ocaml-multicore/dscheck/pull/26) that is available on OCaml 4.  In addition, this ports the STM tests to build on OCaml 4 and use `STM_thread` when possible.  All of this allows [OCaml-CI](https://github.com/ocurrent/ocaml-ci) to build and run tests on OCaml 4.  This allows to remove the two GH action based builds on OCaml 4 (see #116).  All of this should help to make sure this package works on OCaml 4 as intended.


saturn_lockfree.opam.template
```diff
@@ -0,0 +1,3 @@
+pin-depends: [
+  [ "dscheck.dev" "git+https://github.com/ocaml-multicore/dscheck#c81177e2f17d8ca336e14b47251b4473b393e750" ]
```
polytypic, line 2: A new dscheck version must be released and this pin-depend removed before merging.

bench/bench_queue.ml
```diff
@@ -46,8 +46,13 @@ let run_one ~budgetf ?(n_adders = 2) ?(n_takers = 2)
         let n = Util.alloc n_msgs_to_take in
         if n <> 0 then
           let rec loop n =
-            if 0 < n then
-              loop (n - Bool.to_int (Option.is_some (Queue.pop_opt t)))
+            if 0 < n then begin
+              match Queue.pop_opt t with
+              | None ->
+                  Domain.cpu_relax ();
```
polytypic, line 52: `Domain.cpu_relax ()` on OCaml 4 translates to calling `Thread.yield ()` via `domain_shims`.  In older versions of OCaml 4 poll points are not automatically added to all loops.  This means that the emulation of domains by `domain_shims` may break down as a loop that never polls cannot be stopped by the runtime and the program gets stuck.  So, this PR contains a lot of additions of `Domain.cpu_relax ()` calls to loops.  Some of those calls might be unnecessary, but most should be relatively harmless.  We can also remove unnecessary calls, if any, later.

test/skiplist/qcheck_skiplist.ml
```diff
@@ -119,32 +111,29 @@ let tests_two_domains =
...
+            Domain.spawn @@ fun () ->
+            Barrier.await barrier;
+            work lpush1
+          in
+          let () =
+            Barrier.await barrier;
             work lpush2
           in
-          let _ = Domain.join domain1 in
+          let () = Domain.join domain1 in
```
polytypic, line 136: We should avoid `let _ = ...`.

test/ws_deque/stm_ws_deque.ml
```diff
@@ -60,29 +60,30 @@ module WSDConf = struct
...
-      repeat rep_count WSDT_dom.agree_prop_par_asym triple)
-
 let () =
-  let count = 1000 in
-  QCheck_base_runner.run_tests_main
+  let make_domain ~count ~name
+      (module Dom : Stm_run.STM_domain
+        with type Spec.cmd = Spec.cmd
+         and type Spec.state = Spec.state
+         and type Spec.sut = Spec.sut) =
```
polytypic, line 68: This is a case where first-class modules are a bit cumbersome in OCaml, because the syntax for constraints is restricted (i.e. can't specify `(module Dom : STM_run.STM_domain with module Spec = Spec)`), but, on the whole, first-class module perhaps makes things a bit more concise.

.github/workflows/main.yml
```diff
@@ -46,32 +46,3 @@ jobs:
 
       - name: Test
         run: opam exec -- dune runtest
-
-  build-4_x:
```
polytypic, line 50: There is no longer need to use GH actions to test building on OCaml 4.

lyrm (approved):
Thanks for the PR !

PR #129, opened 2024-03-01
Fix MS queue test loop loop without safe point

polytypic:
I noticed a test was hanging in the MS queue optimization PR and I found another test loop that didn't have the necessary safe point for OCaml 4 to allow switching between threads.  This fixes that by adding a `Domain.cpu_relax ()` to the loop, which translates to `Thread.yield ()` on OCaml 4 via the domain shims.

### DIFF
--- a/test/michael_scott_queue/qcheck_michael_scott_queue.ml
+++ b/test/michael_scott_queue/qcheck_michael_scott_queue.ml
@@ -1,4 +1,4 @@
-open Saturn_lockfree.Queue
+module Queue = Saturn_lockfree.Queue
 
 let tests_sequential =
   QCheck.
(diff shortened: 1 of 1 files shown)


test/michael_scott_queue/qcheck_michael_scott_queue.ml
```diff
@@ -1,4 +1,4 @@
-open Saturn_lockfree.Queue
+module Queue = Saturn_lockfree.Queue
```
polytypic, line 1: Let's not use `open` on the data structure we are about to test.  `open`ing the data structure modules makes it more difficult to identify direct uses of the data structure, which is pretty important when reading these tests.

test/michael_scott_queue/qcheck_michael_scott_queue.ml
```diff
@@ -73,28 +75,31 @@ let tests_one_consumer_one_producer =
...
               (fun acc item ->
-                let popped = ref None in
-                while Option.is_none !popped do
-                  popped := pop_opt queue
-                done;
-                acc && item = Option.get !popped)
+                let rec pop_one () =
+                  match Queue.pop_opt queue with
+                  | None ->
+                      Domain.cpu_relax ();
```
polytypic, line 95: This is the place where safe/poll point was missing.

PR #133, opened 2024-05-15
Optimize spsc queue : optimizations without `Obj.magic`

lyrm:
This PR regroups the "safe" optimization implemented for the spsc queue  in #121 by @polytypic. Safe here means that they do not introduce new calls to  'Obj.magic'  or related unsafe OCaml features (like `external ... "%identity"`).

This PR includes the following optimizations: 
- caches of the other side index 
- avoid wide pattern in record, assuming CSE instead
- drop `mask` field 
- micro-optimize empty/full checks 
- avoid code duplication with GADTs
- use of `Atomic.make_contended` if OCaml version >= 5.2.0

This results in a performance increase of a factor 4 to 5 for a large queue.  

old_VS_safe


src_lockfree/spsc_queue.ml
```diff
@@ -24,82 +24,79 @@
...
+    t.head_cache := head;
+    head == head_cache
+  then match mono with Unit -> raise Full | Bool -> false
+  else begin
+    Array.unsafe_set t.array (tail land (size - 1)) (Some element);
+    Atomic.incr t.tail;
+    match mono with Unit -> () | Bool -> true
+  end
+
+let push t element = push_as t element Unit
```
polytypic, line 64: This wasn't changed/introduced in this PR, but I wonder if we should start using the convention that an operation like this raises an exception to indicate an expected condition (rather than an error) would be named with an `_exn` suffix.  IOW, `pop -> pop_exn` and `push -> push_exn`.

src_lockfree/spsc_queue.ml
```diff
@@ -24,82 +24,79 @@
...
+  let size = Array.length t.array in
+  let tail = Atomic.get t.tail in
+  let head_cache = !(t.head_cache) in
+  if
+    head_cache == tail - size
+    &&
+    let head = Atomic.get t.head in
+    t.head_cache := head;
+    head == head_cache
+  then match mono with Unit -> raise Full | Bool -> false
```
polytypic, line 57: I think that we could use `raise_notrace` for exceptions like this which are used primarily for control flow.

Usually one would use a pattern like this with `_exn` functions:

```ocaml
match Q.pop_exn q with
| value -> (* ... *)
| exception Empty -> (* ... *)
```

Maybe we should document this also, i.e. the idea is that `Empty` here is mostly for control flow to avoid allocations and that one should typically use a match like above?

src_lockfree/spsc_queue.ml
```diff
@@ -24,82 +24,79 @@
...
-    array = Array.init size (fun _ -> None);
-  }
+  if size_exponent < 0 || Sys.int_size - 2 < size_exponent then
+    invalid_arg "size_exponent out of range";
+  let size = 1 lsl size_exponent in
+  let array = Array.make size None in
+  let tail = Atomic.make_contended 0 in
+  let tail_cache = ref 0 in
+  let head = Atomic.make_contended 0 in
+  let head_cache = ref 0 in
```
polytypic, line 42: The cached index refs, `tail_cache` and `head_cache`, will potentially suffer from false sharing.

The cache indices do not need to be stored as atomics, but you could change these to be contended atomics, for example.  It would likely improve performance under contention.  (I actually checked and this improved performance on my laptop.) Alternatively you could otherwise allocated the cached indices to avoid false sharing.  For example, if you would allocate a `int array` of correct size,

```ocaml
let tail_cache = Array.make (Obj.size (Obj.repr tail)) 0
```

then that would avoid the false sharing with the current runtime.  (You would then use the index 0 of the array as the cached index value.)

polytypic:
For reference, here is a C++ implementation of a SPSC queue:

```c++
#include <atomic>
#include <cassert>
#include <iostream>
#include <thread>

struct end {
  std::atomic<size_t> m_index;
  size_t m_other_index;

  end() : m_index(0), m_other_index(0) {}
};

template <class T> struct spsc {
  size_t m_mask;
  T *m_array;
  alignas(64) end m_lo;
  alignas(64) end m_hi;

  spsc(size_t capacity) {
    assert((capacity & (capacity - 1)) == 0);

    m_mask = capacity - 1;
    m_array = new T[capacity];
  }

  ~spsc() { delete[] m_array; }

  bool try_push(const T &value) {
    size_t hi = m_hi.m_index.load(std::memory_order_relaxed);
    if (hi - m_hi.m_other_index <= m_mask) {
      m_array[hi & m_mask] = value;
      m_hi.m_index.store(hi + 1, std::memory_order_release);
      return true;
    }
    m_hi.m_other_index = m_lo.m_index.load(std::memory_order_acquire);
    if (hi - m_hi.m_other_index <= m_mask) {
      m_array[hi & m_mask] = value;
      m_hi.m_index.store(hi + 1, std::memory_order_release);
      return true;
    }
    return false;
  }

  bool try_pop(T &out) {
    size_t lo = m_lo.m_index.load(std::memory_order_relaxed);
    if (0 < (m_lo.m_other_index - lo)) {
      out = m_array[lo & m_mask];
      m_lo.m_index.store(lo + 1, std::memory_order_release);
      return true;
    }
    m_lo.m_other_index = m_hi.m_index.load(std::memory_order_acquire);
    if (0 < (m_lo.m_other_index - lo)) {
      out = m_array[lo & m_mask];
      m_lo.m_index.store(lo + 1, std::memory_order_release);
      return true;
    }
    return false;
  }

  spsc(const spsc &) = delete;
  spsc &operator=(const spsc &) = delete;
};

int main(int argc, char *argv[]) {
  spsc<size_t> queue(512);

  const size_t n_messages = 1000 * 1000 * 500;

  std::thread producer([&]() {
    size_t n = n_messages;
    while (0 < n) {
      if (queue.try_push(n))
        --n;
    }
  });

  size_t n = n_messages;
  while (0 < n) {
    size_t message;
    if (queue.try_pop(message))
      --n;
  }

  producer.join();

  return 0;
}
```

On my laptop this seems to run in about a second, which translates to 500 million messages per second.

```sh
➜  clang++ -O2 -std=c++17 spsc.cpp && hyperfine -N --warmup 1 ./a.out
Benchmark 1: ./a.out
  Time (mean ± σ):     990.8 ms ±   6.0 ms    [User: 1976.3 ms, System: 1.2 ms]
  Range (min … max):   981.2 ms … 999.5 ms    10 runs
```

This is not a production quality implementation (I didn't test this extensively &mdash; there might be bugs; this doesn't guard against some obvious programming errors) and I didn't spend a lot of time tweaking this &mdash;  I would have actually assumed to reach closer to 1 billion message per second on my laptop.  I did look at the generated code on godbolt and didn't spot any obvious significant issues with the generated code.

polytypic:
So, I wanted to show the C++ SPSC implementation for a specific reason.  When optimizing code, it is important to understand the context.  For example, if you are optimizing an application then you might have a goal such as "make a specific operation in the application run in some specific time period".  For example, it might take 0.5 seconds for an operation in the application.  That is long enough that a person will notice that something is not instantenous.  A goal could then be to optimize that operation to run in less than 0.25 seconds, [which is typically fast enough that a person does not notice that something is taking a lot of time](https://www.pubnub.com/blog/how-fast-is-realtime-human-perception-and-technology/).  If and once you make the operation run in 0.25 you are (or may be) done.  You don't have to think about whether things could still be faster.

Unfortunately, when implementing a library, we do not know the requirements of any particular application.  What should we compare against?  The way I think about optimizing libraries like this is that I want to understand what is the (estimated) _**theoretical peak performance**_ that could be achieved.  There is no single way to figure out what is the theoretical peak.  You could, for example, consider things like maximum memory bandwidth and cost of atomic (fenced) instructions, a single threaded reference implementation (and multiply by two for 2 threads working mostly without interference), a reference implementation in C++.  The way I then judge my optimizations is with respect to the theoretical peak performance I came up with.  I want to get as close as possible to the theoretical peak performance &mdash; and if I don't get close enough, I spend time to figure out why (is my estimated theoretical peak too optimistic? is something in my implementation working poorly?).  Depending on what exactly one is working on, getting to within some percentage of theoretical peak could then be acceptable &mdash; for something like this, I'd want to get within maybe 90%-95% of what is possible with OCaml and perhaps within 2x of what is possible with a language like C++.  I basically want things to so fast that almost nobody should feel the urge to write their own implementation to optimize their application.

polytypic:
Optimizing these kinds of data structures is all about optimizing for the cache architecture and memory bandwidth.

For example, this implementation uses 3x more words per memory due to the `_ option array` than the unsafe implementation using a non-option array.  This basically means is that if you estimate the theoretical peak performance based on memory bandwidth, then you would expect this to perform up to 3x slower in cases when you are bandwidth bound.  Try to, for example, make the queue very large (large size exponent) and pass a lot of messages through the queue. Quite possibly at some point you will be bound by memory bandwidth and this queue will then likely be about factor of three slower than the unsafely optimized queue.  This is basically a physical law of nature.  It simply takes 3x more time to write/read 3x more memory.  There is no way around that.

When you are not bound by memory bandwidth, you are likely bound by the cache architecture.  On a modern CPU every write to memory by CPU core effectively "takes ownership" of the cache line aligned memory region affected by the write.  This happens via the cache coherency protocol, i.e. after a write the L1 cache of the core that performed the write "owns" the region of memory.  Another CPU core accessing the same region will then take a performance hit when trying to access (read or write) that same cache line aligned region of memory.  This is the reason to avoid false sharing where a single cache line aligned region of memory is being written to by at least one CPU core and also accessed by other cores (but not accessing the exact same words).  The cached indices and the queue record in this version of the queue are not protected against false sharing.  Sometimes you get lucky and things run fast.  Other times there is false sharing/contention and things run much slower.  For an application that has any sort of real-time requirements can be very undesirable.

polytypic:
Personally I think that we should "get a blessing" to at least use the equivalent of `Multicore_magic.copy_as_padded`.  I hope there is a way to write such a function in a way that it is guaranteed to work even with a compiler Flambda.  Otherwise we should actually warn ⚠️  that the **_safe implementation are unsafe against false sharing_** &mdash; that is, their performance can sometimes be dramatically worse due to false sharing.

The factor of 3.5 slowdown is actually a lucky case.  I would assume that I could actually make that way worse by writing a program that has a falsely shared mutable integer field with the queue that the program would be writing to from a thread that gets allocated to a different core cluster.  Then the cache line transfer would need to go between core clusters, which is likely much slower than within a core cluster.  It might also be that if the other core cluster is an "efficiency" cluster (i.e. the cores are slower in such a cluster) the transfer cost would be even higher.

polytypic:
In summary:
* I think we should use the `_exn` naming pattern and `raise_notrace` for operations that raise exceptions mostly as a control flow mechanism to avoid memory allocations.  (I actually benchmarked that with `raise_notrace` using exceptions for control flow was slightly faster on my laptop than returning an option.)
* I'd strongly recommend picking some way to explicitly guard even safe implementations agaist false sharing.  I think the simplest way to do that would be to get a "blessed" implementation of `copy_as_padded`.  Othewise I would recommend labelling safe data structure implementations as ⚠️  **_unsafe against false sharing_**.  Doing this will also not only make performance more stable, but will likely also improve thruput in most cases.

Otherwise this PR looks good to me. 👍

polytypic:
> This method can be used by at most one domain at a time.

I noticed this wording in the documentation.  I've actually been asked about these in a private communication.  I think we should clarify our definition of "single consumer" or "single producer".

The comment implies a too restrective meaning.  It is perfectly fine for a single consumer/producer to move from one domain or (sys)thread to another.  What we want, informally, is that actions on the queue from a single consumer/producer are ordered as if they would come from a single thread of execution.  It is not even required that a single producer/consumer would be strictly speaking a single thread of execution.

Perhaps we should say something like that producer (and consumer) actions must be totally ordered according to the [happens-before relation](https://v2.ocaml.org/manual/memorymodel.html#s:happens_before) defined by the OCaml memory model.

polytypic:
With the padding of the `_cache` indices, performance in the benchmark is now more stable and improved on my laptop:

```sh
➜  saturn git:(optimize-spsc-queue-safe) ✗ dune exec --release -- bench/main.exe -budget 1 -brief Single
Saturn_lockfree Single_prod_single_cons_queue:
  time per message/2 workers, capacity 1:
    345.68 ns
  messages over time/2 workers, capacity 1:
    5.79 M/s
  time per message/2 workers, capacity 8:
    134.05 ns
  messages over time/2 workers, capacity 8:
    14.92 M/s
  time per message/2 workers, capacity 64:
    66.87 ns
  messages over time/2 workers, capacity 64:
    29.91 M/s
  time per message/2 workers, capacity 512:
    55.23 ns
  messages over time/2 workers, capacity 512:
    36.22 M/s
  time per message/2 workers, capacity 4096:
    40.16 ns
  messages over time/2 workers, capacity 4096:
    49.80 M/s
  time per message/2 workers, capacity 32768:
    21.65 ns
  messages over time/2 workers, capacity 32768:
    92.38 M/s
```

Because the main record of the queue is not padded, it is still possible for the queue to suffer from false sharing, but this does not show in our benchmark, because the benchmark is written so that no false sharing could really happen.  A real program could still easily accidentally get false sharing on the main queue record.

polytypic (approved):

PR #135, opened 2024-05-24
Optimize spsc queue : unsafe optimizations

lyrm:
This PR follows PR #133  and adds a new spsc queue using the safely optimized spsc queue augmented with the  PR #121  unsafe optimizations made by @polytypic.   

This includes : 
    - Add more padding to avoid false sharing using `Multicore_magic.copy_as_padded` 
    - Use `Atomic = Transparent_atomic`, `fenceless_get`, and `incr`
    - Avoid indirection in array  (removing the use of the option type)
    - Avoid float array pessimization
    - Add a simple `float` test and prevent inlining

I will add some benchmarks later here, but I am interested on knowing the performance you got on yours, @polytypic .


polytypic:
```sh
➜  saturn git:(optimize-spsc-queue-unsafe) ✗ dune exec --release -- bench/main.exe -budget 1 -brief Single
Saturn_lockfree Single_prod_single_cons_queue:
  time per message/2 workers, capacity 1:
    338.69 ns
  messages over time/2 workers, capacity 1:
    5.91 M/s
  time per message/2 workers, capacity 8:
    120.25 ns
  messages over time/2 workers, capacity 8:
    16.63 M/s
  time per message/2 workers, capacity 64:
    63.43 ns
  messages over time/2 workers, capacity 64:
    31.53 M/s
  time per message/2 workers, capacity 512:
    55.28 ns
  messages over time/2 workers, capacity 512:
    36.18 M/s
  time per message/2 workers, capacity 4096:
    39.26 ns
  messages over time/2 workers, capacity 4096:
    50.94 M/s
  time per message/2 workers, capacity 32768:
    19.05 ns
  messages over time/2 workers, capacity 32768:
    104.98 M/s
Saturn_lockfree Single_prod_single_cons_queue_unsafe:
  time per message/2 workers, capacity 1:
    295.25 ns
  messages over time/2 workers, capacity 1:
    6.77 M/s
  time per message/2 workers, capacity 8:
    118.22 ns
  messages over time/2 workers, capacity 8:
    16.92 M/s
  time per message/2 workers, capacity 64:
    47.08 ns
  messages over time/2 workers, capacity 64:
    42.48 M/s
  time per message/2 workers, capacity 512:
    47.76 ns
  messages over time/2 workers, capacity 512:
    41.88 M/s
  time per message/2 workers, capacity 4096:
    29.54 ns
  messages over time/2 workers, capacity 4096:
    67.71 M/s
  time per message/2 workers, capacity 32768:
    8.11 ns
  messages over time/2 workers, capacity 32768:
    246.53 M/s
```

lyrm:
Good idea for the benchmarks, I would do that.

src_lockfree/spsc_queue_unsafe.ml
```diff
@@ -0,0 +1,111 @@
...
+      | Pop ->
+          Array.unsafe_set t.array index (Obj.magic ());
+          Atomic.incr t.head
+      | Peek -> ()
+    end;
+    match poly with Value -> v | Option -> Some v
+
+let pop t = pop_or_peek_as t Pop Value
+let pop_opt t = pop_or_peek_as t Pop Option
+let peek t = pop_or_peek_as t Peek Value
```
polytypic, line 105: We should also rename `pop -> pop_exn`, `peek -> peek_exn`, and `push -> push_exn` to match the safe version.

src_lockfree/spsc_queue_unsafe.ml
```diff
@@ -0,0 +1,111 @@
...
+  let size = Array.length t.array in
+  let tail = Atomic.fenceless_get t.tail in
+  let head_cache = !(t.head_cache) in
+  if
+    head_cache == tail - size
+    &&
+    let head = Atomic.get t.head in
+    t.head_cache := head;
+    head == head_cache
+  then match mono with Unit -> raise Full | Bool -> false
```
polytypic, line 66: We should also use `raise_notrace` like in the safe version (and document that we do so).

We could extract the common signature to a separate `spsc_queue_intf.ml` file to avoid having to duplicate documentation.

test/spsc_queue/test_spsc_queue.ml
```diff
@@ -31,7 +31,7 @@ let test_parallel () =
   let producer =
     Domain.spawn (fun () ->
         for i = 1 to count do
-          while not (push_not_full q i) do
+          while not (push_not_full q (Float.of_int i)) do
```
polytypic, line 34: I think we should either functorize the tests or maybe use some other tricks to test both the safe and unsafe versions.

src_lockfree/spsc_queue_unsafe.ml
lyrm: I did the common signature and applied this comment and the previous ones. Thanks !

test/spsc_queue/test_spsc_queue.ml
lyrm: I pushed a solution to run all the tests with both the safe and the unsafe implementations ! Thanks for your review.

bench/bench_spsc_queue.ml
```diff
@@ -1,46 +1,96 @@
...
+          loop n_msgs
       in
-      loop n_msgs
+      (work, before)
   in
 
-  let config = Printf.sprintf "2 workers, capacity %d" (1 lsl size_exponent) in
+  let config =
+    Printf.sprintf "2 workers, capacity %d (%s)" (1 lsl size_exponent)
+      (if unsafe then "unsafe" else "safe")
```
polytypic, line 83: Hmm... So, this will add `(safe)` or `(unsafe)` as a suffix.

How about just adding the `(unsafe)` suffix and not adding a suffix for "safe" versions?

This would be nice in two ways:
1. The existing benchmark does not have the `(safe)` suffix.  So, keeping the existing naming means that there is no discontinuity in the benchmark graphs.
2. If, at some point, OCaml is enhanced to avoid the need for the unsafe versions, we can then just drop the unsafe version and don't have to rename the safe versions, because at that point a `(safe)` suffix would be superfluous.

lyrm: Good point, thanks !

polytypic (approved):
Other than the `(safe)` and `(unsafe)` tagging this LGTM.  🎉

PR #137, opened 2024-05-25
Reintroduce dropped commits to safe spsc PR

polytypic:
These changes that I pushed to the #133 PR branch were dropped for some reason when the PR was first squashed and then merged.


lyrm (approved):
Sorry for my mistake and thanks for this PR !

PR #139, opened 2024-05-28
Require `sherlodoc` for search bar

polytypic:
This requires `dune` 3.14, `odoc` 2.4.1, and `sherlodoc` 0.2 so that generated documentation should include the [new search bar](https://discuss.ocaml.org/t/ann-dune-3-14/14096).

Make sure to install `sherlodoc` when making a release and updating the generated documentation!


lyrm (approved):
Thanks for the PR !

PR #140, opened 2024-05-29
Fix cycle dependency in skiplist tests

lyrm:
The skiplist tests required the `saturn` library while being part of the `saturn_lockfree` package, likely creating a circular dependency and preventing successful installation when running opam install . --with-test. This patch resolves that issue.


polytypic (approved):
Thanks!

PR #143, opened 2024-07-02
Using multicore-magic-dscheck for dscheck tests

lyrm:
This PR fixes the issue with running DScheck tests on data structures using multicore-magic `Transparent_atomic` and/or `Atomic_array` modules. They both need do have their atomic calls shadowed by DScheck's atomics and until now, we were using some working but not generic tricks. Especially, this had some visible consequences on the saturn library itself (with a `transparent_atomic` module that only had that use).  

TODO : 
- [ ]  Add the proper dependencies to multicore-magic-dscheck and DScheck when  their releases.


polytypic:
This looks nice! 👍

polytypic (approved):
We obviously need to get rid of the `pin-depends`, but otherwise this looks very good. So, I'm marking this as approved. 👍

It will probably take some time to go through the opam publish manoeuvers twice (for dscheck and multicore-magic).

PR #146, opened 2024-07-11
Michael-Scott queue : safe - unsafe versions

lyrm:
This PR follows PR #122 work: the optimizations or changes made in this previous PR are separated into two to get a safe Michael-Scott queue and an "unsafe" one (with Obj.magic use). 

  
TODO : 
- [x] For now, the `dscheck` tests for  the unsafe version get segmentation faults.   
- [x] Documentation


lyrm:
@polytypic: The dscheck tests of the unsafe version result in a segmentation fault. I encountered the same issue with PR #122 after rebasing on main and adding the few necessary changes. Could you have a look?

lyrm:
@polytypic  : this PR is ready for review. Could you have a look ?

README.md
```diff
@@ -286,6 +304,16 @@ Because of the great properties of OCaml 5 memory model (see the
 more details), not a lot can go wrong here. At least, data corruption or
 segmentation fault won't happen like it can in other languages.
 
+## Safe and unsafe data structures
+Some data structures are available in two versions: a normal version and a more optimized but **unsafe** version. The **unsafe** version utilizes `Obj.magic` in a way that may be unsafe with `flambda2` optimizations.
```
polytypic, line 308: Let's format .md files with prettier.  I recommend setting up your editor so that it does that automatically.

README.md
```diff
@@ -33,15 +34,24 @@ is distributed under the
 
 # Contents
 
+- [Saturn — Parallelism-Safe Data Structures for Multicore OCaml](#saturn--parallelism-safe-data-structures-for-multicore-ocaml)
```
polytypic, line 37: This is not an issue with this PR, but some time ago GitHub started to automatically produce a kind contents or navigation drop down for .md files.  So, I've been thinking of potentially starting to drop contents section from .md files (less stuff to maintain).  On quick look ocaml.org/packages doesn't yet produce such a navigation helper, so I guess there is still value in having a contents section.  So, just mention this here as a potential thing to discuss.

src_lockfree/michael_scott_queue.ml
```diff
@@ -75,7 +72,7 @@ let rec fix_tail tail new_tail =
     && not (Atomic.compare_and_set tail old_tail new_tail)
   then fix_tail tail new_tail
 
-let push { tail; _ } value =
+let push_exn { tail; _ } value =
```
polytypic, line 75: Hmm... The convention has been to use the `_exn` suffix on operations that raise an expected exception in some normal cases (such as when the queue is empty or full), but in this case no such exception is being used.  Why should we use the `_exn` suffix in this case?

README.md
lyrm: It should have been working this way, but apparently, my editor config was not working right ! Thanks for catching it.

src_lockfree/michael_scott_queue.ml
lyrm: Good catch ! I made the change on `saturn_benchmarks` to match the generic `Queue` signature.  I should not have move it here !

polytypic:
Looking at the "safe" version, the push logic:

```ocaml
(* ... *)

let push { tail; _ } value =
  let rec find_tail_and_enq curr_end node =
    if not (Atomic.compare_and_set curr_end Nil node) then
      match Atomic.get curr_end with
      | Nil -> find_tail_and_enq curr_end node
      | Next (_, n) -> find_tail_and_enq n node
  in
  let new_tail = Atomic.make Nil in
  let newnode = Next (value, new_tail) in
  find_tail_and_enq old_tail newnode;
  if not (Atomic.compare_and_set tail old_tail new_tail) then
    fix_tail tail new_tail
```

doesn't have some of the (entirely safe) optimizations from the "unsafe" version:

```ocaml
(* ... *)

let push { tail; _ } value =
  let (Next _ as new_node : (_, [ `Next ]) Node.t) = Node.make value in
  let link = Node.as_atomic old_tail in
  if Atomic.compare_and_set link Nil new_node then
    Atomic.compare_and_set tail old_tail new_node |> ignore
  else
    let backoff = Backoff.once Backoff.default in
    push tail link new_node backoff
```

In particular, notice how the "unsafe" version does not call any recursive functions in the fast path:

```ocaml
let push { tail; _ } value =
  let (Next _ as new_node : (_, [ `Next ]) Node.t) = Node.make value in
  let link = Node.as_atomic old_tail in
  (* first we try the fast path *)
  if Atomic.compare_and_set link Nil new_node then
    Atomic.compare_and_set tail old_tail new_node |> ignore
  else
    (* failed, so we do the slow path *)
```

It should be possible to incorporate this optimization into the "safe" version and it might speed it up a tiny bit.

test/michael_scott_queue/michael_scott_queue_dscheck.ml
```diff
@@ -1,151 +1,165 @@
...
           test_case "2-domains" `Slow two_domains;
         ] );
     ]
+end
+
+let () =
+  let module Safe = Dscheck_ms_queue (Michael_scott_queue) in
+  let safe_test = Safe.tests "safe" in
+  let module Unsafe = Dscheck_ms_queue (Michael_scott_queue_unsafe) in
+  let unsafe_test = Unsafe.tests "unsafe" in
```
polytypic, line 162: Not a major issue, but this would probably be a bit less verbose with first-class modules.  It could be like:

```ocaml
let safe_tests = tests "safe" (module Michael_scott_queue) in
let unsafe_tests = tests "unsafe" (module Michael_scott_queue_unsafe) in
...
```

The module language can be a bit cumbersome at times and in a case like this where we are only interested in a simple (non module) return value (a list of tests) first-class modules can be (IMHO) more convenient.

test/michael_scott_queue/stm_michael_scott_queue.ml
```diff
@@ -1,65 +1,73 @@
...
-  Stm_run.run ~count:500 ~verbose:true
-    ~name:"Saturn_lockfree.Michael_scott_queue"
-    (module Spec)
-  |> exit
+  let module Safe = STM_ms_queue (Ms_queues.Michael_scott_queue) in
+  let exit_code = Safe.run () in
+  if exit_code <> 0 then exit exit_code
+  else
+    let module Unsafe = STM_ms_queue (Ms_queues.Michael_scott_queue_unsafe) in
+    Unsafe.run () |> exit
```
polytypic, line 73: This could also be a bit less verbose using first-class modules.  We are not really interested in the `Safe` and `Unsafe` modules as such.  We just want to run the tests and get an int as a result.

lyrm:
About the missing safe optimizations: I added the fast path in the push function. However, the benchmarks suggest that adding backoff (either for in `fix_tail` or `find_tail_and_end`) does not improve performance (at least on my computer). Did I miss other safe optimizations ?

polytypic:
> About the missing safe optimizations: I added the fast path in the push function. However, the benchmarks suggest that adding backoff (either for in `fix_tail` or `find_tail_and_end`) does not improve performance (at least on my computer). Did I miss other safe optimizations ?

Hmm... One could also leave the backoff out.  It is possible that the backoff helps on other CPU (micro)architectures and we can add that later.

polytypic (approved):
The logic looks good to me. 🎉

toots:
Hi there! While I understand and appreciate the hard work that went into optimizing the queue, what is the recommended alternative to grab the elements of the queue? A `pop` loop followed by a `push` loop?

polytypic:
As a side note, the removal of the snapshot mechanism wasn't just about optimization.

Here is a puzzle about the "snapshot" mechanism.  The code below uses the previous 0.4.1 version of this library, which is the only published version of this library that has the MS queue with the "snapshot" mechanism and no (additional) space leaks.  Let's first define a module alias:

``` ocaml
module Q = Saturn.Queue
```

Here is a function that converts a "snapshot" into a list:

```ocaml
let[@tail_mod_cons] rec to_list s =
  match Q.next s with
  | None -> []
  | Some (v, s) -> v :: to_list s
```

Question A: What is the return value of the below program?

```ocaml
let q = Q.create () in
let s = Q.snapshot q in
Q.push q 101;
ignore (Q.pop q);
to_list s
```

Question B: What is the return value of the below program?

```ocaml
let q = Q.create () in
Q.push q 101;
let s = Q.snapshot q in
Q.push q 42;
ignore (Q.pop q);
ignore (Q.pop q);
to_list s
```

<details>
  <summary>Click to see answer and explanation</summary>

 
The first program returns `[]` and the second program returns `[101; 42]`.


To me, a "snapshot" should be an immutable instant. But that is not how the mechanism worked.  Instead, `snapshot` simply returned a pointer to the first non-dummy (if any) node of the queue and the `next` operation then dereferenced the (mutable) pointer to the next node.

This meant that if the queue happened to be empty, the snapshot would be empty. If the queue happened to be non-empty, the snapshot would potentially contain all elements pushed to the queue at that point and all the element to be pushed to the queue in the future.

This behaviour of the snapshot mechanism is, of course, not limited to simple sequential programs.  Concurrent pushes to the queue are also potentially visible through the "snapshot".
</details>

toots:
Good to know! Given the nature of the concurent queue, I was never expecting a strict control over the interleaving calls but this is clearly opposite to what it was advertising.

I think it'd be nice to have a way to get through the elements in the queue without removing them, basically a deeper peek API.

I might see if I can propose a PR for that.

polytypic:
Note that providing a snapshot operation can be easier for some other lock-free queue approaches.

The ["lazy" queue](https://github.com/ocaml-multicore/saturn/pull/145), for example, makes it particularly easy.

Also the approach using [two (essentially immutable) stacks](https://github.com/ocaml-multicore/saturn/pull/112), it is quite easy to take a snapshot of the queue.  The Picos project has an implementation of [a two-stack multi-producer, multi-consumer queue](https://github.com/ocaml-multicore/picos/tree/83bfb1e2aba3810ca35461b1876587a3047cd8e7/lib/picos_mpmcq) that should be relatively easy to extend with a snapshot operation.

Also, [the queue provided with Kcas has snapshot functionality](https://ocaml-multicore.github.io/kcas/doc/kcas_data/Kcas_data/Queue/index.html#val-to_seq).

Issue #56, opened 2023-02-23
Single core benchmarks


lyrm:
This library lacks some single core benchmarks  to compare its domain safe data structures to the standard library data structures (that are not thread safe). This idea is to measure the overhead paid for lock-freedom  and multi-domain safety.

| Lockfree data structures | Standard library data structures     | 
|-------------|------------|
| [Michael-Scott queue]( https://github.com/ocaml-multicore/lockfree/blob/main/src/michael_scott_queue.mli )  | [Stdlib.Queue](https://v2.ocaml.org/api/Queue.html) |
| [Mpmc queue](https://github.com/ocaml-multicore/lockfree/blob/main/src/mpsc_queue.mli) |  [Stdlib.Queue](https://v2.ocaml.org/api/Queue.html) | 
| [Spsc queue](https://github.com/ocaml-multicore/lockfree/blob/main/src/spsc_queue.mli) |  [Stdlib.Queue](https://v2.ocaml.org/api/Queue.html) |
| [Working stealing dequeue](https://github.com/ocaml-multicore/lockfree/blob/main/src/ws_deque.mli) |  [Core.Deque](https://ocaml.janestreet.com/ocaml-core/109.55.00/tmp/core_kernel/Dequeue.html) |
|  [Treiber stack](https://github.com/ocaml-multicore/lockfree/blob/main/src/treiber_stack.mli) | [Stdlib.Stack](https://v2.ocaml.org/api/Stack.html) |

Vincent-lau:
Hi I have run some simple benchmarks by extending the `bench_queue.ml` for comparsion of lock free queues and the standard library queues. Here are the results

```
Saturn_lockfree Queue:
  time per message/one domain:
    75.12 ns
  messages over time/one domain:
    13.31 M/s
  time per message/1 nb adder, 1 nb taker:
    240.00 ns
  messages over time/1 nb adder, 1 nb taker:
    8.33 M/s
  time per message/1 nb adder, 2 nb takers:
    378.39 ns
  messages over time/1 nb adder, 2 nb takers:
    7.93 M/s
  time per message/2 nb adders, 1 nb taker:
    212.92 ns
  messages over time/2 nb adders, 1 nb taker:
    14.09 M/s
  time per message/2 nb adders, 2 nb takers:
    296.55 ns
  messages over time/2 nb adders, 2 nb takers:
    13.49 M/s
Stdlib Queue:
  time per message/one domain:
    27.04 ns
  messages over time/one domain:
[...]

PR #150, opened 2024-09-06
Add benchmark for Stdlib.Queue

Vincent-lau:
For comparsion between Saturn_lockfree Queue and Stdlib.Queue.

### DIFF
--- a/bench/bench_queue.ml
+++ b/bench/bench_queue.ml
@@ -1,19 +1,43 @@
 open Multicore_bench
-module Queue = Saturn_lockfree.Queue
 
-let run_one_domain ~budgetf ?(n_msgs = 50 * Util.iter_factor) () =
-  let t = Queue.create () in
+module type BENCHQUEUE = sig
+  type 'a t
 
-  let op push = if push then Queue.push t 101 else Queue.pop_opt t |> ignore in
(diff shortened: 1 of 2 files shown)


polytypic:
Notice the "Stdlib benchmarks" link at the top of the README:

Screenshot 2024-09-06 at 18 35 32

polytypic:
BTW, one reason the benchmarks do not use functors is that using functors pretty much prevents the native compiler from inlining.  Many of the lock-free data structure implementations have operations that use a worker-wrapper idiom where the wrapper is a trivial function designed to be inlined (often it just calls the recursive worker with an extra parameter or two).

Vincent-lau:
Thanks! So does that mean that issue #56  is actually done by now? Or is there any further benchmarking work needed?

polytypic:
I believe there is planty of benchmarking to be done.  However, it is a complex topic.  Most of the benchmarks we have are high contention benchmarks.  One idea would be to [add delays to simulate work](https://www.youtube.com/watch?v=7HBEXs48qmo&t=1862s).  Also, benchmarks that would exercise both read-only and read-write operations with various data structures could be interesting.

Vincent-lau:
Ok, thanks. Closing this PR now since this work has already been done

PR #154, opened 2024-09-25
Hash table from picos 

lyrm:
Import Picos hash table in Saturn. Some changes: 
- using the size module
- a safe and an unsafe version. Issue : the size module uses `Obj.magic`

TODO :
- [x]  Dscheck tests are taking too long
- [x] the safe and the unsafe implementation are the same except for the use of `Transparent_atomic` and `Multicore_magic.Atomic_array`. Could we avoid duplicating the code somehow ?


bench/bench_htbl.ml
```diff
@@ -0,0 +1,76 @@
...
+    ?(prepopulate = true) (module Htbl : Htbl_intf.HTBL) =
+  let percent_rem = 100 - (percent_mem + percent_add) in
+
+  let limit_mem = percent_mem in
+  let limit_add = percent_mem + percent_add in
+
+  assert (0 <= limit_mem && limit_mem <= 100);
+  assert (limit_mem <= limit_add && limit_add <= 100);
+
+  let t = Htbl.create ~hashed_type:(module Int) () in
```
polytypic, line 14: We are already generating random integer values for the keys, which means that hashing the keys is unnecessary.  We also want to specifically measure the cost of the operations inside the hash table itself and exclude everything else as much as possible, which also includes `Int.hash` (which is also missing from older compiler versions).  So, it would be better to use something like:

```ocaml
module Key = struct
  type t = int

  let equal = Int.equal
  let hash = Fun.id
end
```

for the `hashed_type`.

bench/bench_htbl.ml
```diff
@@ -0,0 +1,76 @@
+open Multicore_bench
+
+let run_one ~budgetf ~n_domains ?(n_ops = 20 * Util.iter_factor)
+    ?(n_keys = 10000) ~percent_mem ?(percent_add = (100 - percent_mem + 1) / 2)
+    ?(prepopulate = true) (module Htbl : Htbl_intf.HTBL) =
```
polytypic, line 5: Hmm... I suspect that in this particular case it might not matter (I'll try to remember to consider this), but the native OCaml compiler is incapable of specializing (inlining) code when using first-class modules or functors.  What I mean here specifically is that in many cases optimized versions of data structures use a worker-wrapper idiom where the wrapper is designed to be inlined.  Using a first-class module or functor, the compiler will not be able to inline the code at all.  This usually tilts the results such that the optimized data structure will apper relatively slower than it would be under normal use when code would be just directly using the optimized version of the data structure (rather than when using it through a first-class module or functor argument like here) and the compiler is able to inline function calls (when it decides to do so).

IOW, in some cases like _benchmarks_, with the OCaml compiler as it is today, it may be better to just duplicate code, because otherwise one might be tilting the results.

src_lockfree/htbl/htbl.ml
```diff
@@ -0,0 +1,645 @@
...
+    }
+      -> ('k, 'v, [> `Cons_with_size ]) tdt
+  | Resize : {
+      spine : ('k, 'v, [ `Nil_with_size | `Cons_with_size ]) tdt;
+    }
+      -> ('k, 'v, [> `Resize ]) tdt
+      (** During resizing and snapshotting target buckets will be initialized
+             with a physically unique [Resize] value and the source buckets will
+             then be gradually updated to [Resize] values and the target buckets
+             updated with data from the source buckets. *)
```
polytypic, line 58: In Emacs you can use `M-x tuareg-indent-phrase` (it is also bound to `M-q` in my Emacs) to reformat comments like this to be aligned nicely.

test/michael_scott_queue/stm_michael_scott_queue.ml
```diff
@@ -1,9 +1,8 @@
+open QCheck
```
polytypic, line 1: These changes should not be in this PR.   It is good practise to always rebase PRs against the main.  It saves time from reviewers.

src_lockfree/htbl/htbl_unsafe.ml
```diff
@@ -0,0 +1,634 @@
...
+   REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF MERCHANTABILITY
+   AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR ANY SPECIAL, DIRECT,
+   INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES WHATSOEVER RESULTING FROM
+   LOSS OF USE, DATA OR PROFITS, WHETHER IN AN ACTION OF CONTRACT, NEGLIGENCE OR
+   OTHER TORTIOUS ACTION, ARISING OUT OF OR IN CONNECTION WITH THE USE OR
+   PERFORMANCE OF THIS SOFTWARE. *)
+
+open Htbl_utils
+module Atomic = Multicore_magic.Transparent_atomic
+module Atomic_array = Multicore_magic.Atomic_array
```
polytypic, line 17: Hmm... The diff between the safe and unsafe versions
```diff
➜  htbl git:(picos_htbl) ✗ diff htbl.ml htbl_unsafe.ml
14d13
< open Htbl_utils
16,28c15,17
< module Atomic_array = struct
<   type 'a t = 'a Atomic.t array
< 
<   let[@inline] at (xs : 'a t) i : 'a Atomic.t = Array.get xs i
<   let[@inline] make n v = Array.init n @@ fun _ -> Atomic.make v
< 
<   external length : 'a array -> int = "%array_length"
< 
<   let unsafe_fenceless_get xs i = Atomic.get xs.(i)
< 
<   let[@inline] unsafe_compare_and_set xs i b a =
<     Atomic.compare_and_set (at xs i) b a
< end
---
> open Htbl_utils
> module Atomic = Multicore_magic.Transparent_atomic
> module Atomic_array = Multicore_magic.Atomic_array
```
seems like something that could probably be handled with a bit of dune build magic.

polytypic: Maybe something along the lines of (this is untested, just looking at [actions](https://dune.readthedocs.io/en/stable/reference/actions/index.html)):

```
(rule 
 (action
  (with-stdout-to htbl.ml
   (progn
    (cat htbl.head_safe.ml)
    (cat htbl.body.ml)))))

(rule 
 (action
  (with-stdout-to htbl_unsafe.ml
   (progn
    (cat htbl.head_unsafe.ml)
    (cat htbl.body.ml)))))
```

src_lockfree/htbl/htbl_unsafe.ml
```diff
@@ -0,0 +1,634 @@
...
+        if r.equal cons_r.key key then
+          Cons { key = cons_r.key; value = cons_r.value; rest = cons_r.rest }
+        else assoc_node r.equal key cons_r.rest
+    | B (Resize resize_r) ->
+        (* A resize is in progress.  The spine of the resize still holds what was
+           in the bucket before resize reached that bucket. *)
+        loop (B resize_r.spine)
+  in
+
+  loop @@ Atomic_array.unsafe_fenceless_get r.buckets i
```
polytypic, line 398: Having an inner function like this with the native compiler generates a heap allocation for a closure.

I have this in `dune-workspace` at the root of project:

```
(lang dune 3.9)

(env
 (dev
  (ocamlopt_flags
   (:standard -S))
  (flags
   (:standard -warn-error -A)))
 (release
  (ocamlopt_flags
   (:standard -S))))
```

This instructs the compiler to produce assembly output (`*.s` files) that you can then find under the `_build` directory hierarchy.

Looking at the generated code (this is actually from the safe version), I can see the following (I've trimmed the output a bit):

```
_camlSaturn_lockfree__Htbl.find_node_926:
;; [SNIP]
	ldr	x16, [x28, #0]  ;; This starts the allocation
	sub	x27, x27, #40
	cmp	x27, x16
	b.lo	L404                ;; This jumps to do GC if necessary (see the last two lines)
L403:	add	x12, x27, #8
	str	x12, [sp, #0]
	movz	x13, #4343, lsl #0
	str	x13, [x12, #-8]
	adrp	x14, _camlSaturn_lockfree__Htbl.loop_934@GOTPAGE
	ldr	x14, [x14, _camlSaturn_lockfree__Htbl.loop_934@GOTPAGEOFF]
	str	x14, [x12, #0]
	movz	x15, #5, lsl #0
	movk	x15, #256, lsl #48
	str	x15, [x12, #8]
	ldr	x22, [sp, #8]
	str	x22, [x12, #16]
	str	x21, [x12, #24]
	mov	x0, x6
	.loc	1	409
	bl	_camlSaturn_lockfree__Htbl.unsafe_fenceless_get_376
L405:
	ldr	x1, [sp, #0]
	.loc	1	409
	ldr	x30, [sp, #24]
	add	sp, sp, #32
	.cfi_adjust_cfa_offset	-32
	b	_camlSaturn_lockfree__Htbl.loop_934
	.cfi_adjust_cfa_offset	32
L404:	bl	_caml_call_gc
L402:	b	L403
```

Doing allocations for `find_node` is unnecessary and is bad for performance.

I would recommend moving the loop to a top level function ([like in the original](https://github.com/ocaml-multicore/picos/blob/849dea9f61319265d947ab4c5c9d134cf3116a74/lib/picos_aux.htbl/picos_aux_htbl.ml#L346-L364)).

test/htbl/stm_htbl.ml
```diff
@@ -0,0 +1,111 @@
...
+  let run () = Stm_run.run ~name:"Htbl" (module Spec) |> exit
+end
+
+let () =
+  let module Safe = STM_htbl (Htbls.Htbl) in
+  let exit_code = Safe.run () in
+  if exit_code <> 0 then exit exit_code
+  else
+    let module Unsafe = STM_htbl (Htbls.Htbl_unsafe) in
+    Unsafe.run () |> exit
```
polytypic, line 111: IMHO, the DSCheck tests take too long &mdash; 361 seconds or 6 minutes on my M3 Max laptop (with 64GB ram).

Unit tests (which is what tests in this repository are) should ideally be something you can comfortably run after every change and should take a matter of seconds rather than minutes.

If we make it so that the safe and unsafe versions are built from essentially the same source file we could just skip testing one of them with DSCheck.  That would drop the time to half (3 minutes), but that is still way too much to run comfortably after every change.

We should find ways to reduce the time further.  One obvious improvement would be to run the DSCheck tests in parallel.  This could be done through dune, because DSCheck itself doesn't support that, but we can run multiple programs in parallel.

bench/bench_htbl.ml
polytypic: Yes, given that both versions use essentially the same code (aside from a short preamble) there is likely not that much difference between the two when it comes to inlining, so in this case using a first-class modules should not tilt the results significantly.

test/htbl/dscheck_htbl.ml
```diff
@@ -0,0 +1,263 @@
...
+
+        Atomic.final (fun () ->
+            Atomic.check (fun () ->
+                (not !added1) && !added2 && (not !added3) && !added4
+                && mem htbl 1 && mem htbl 21 && mem htbl 22)))
+
+  let _two_add_resize () =
+    (* Trying to trigger resize *)
+    let random_offset = Random.int 1000 in
+    for i = 0 to 5 do
```
polytypic, line 57: Hmm... Why do we have the outer loop here? (And in some of the other tests.)

bench/bench_htbl.ml
lyrm: I did not know about that. I can still duplicate the code and add a comment so if the code changes at some points, this is not an issue. What do you think ?

src_lockfree/htbl/htbl_unsafe.ml
lyrm: Works perfectly, thanks !

src_lockfree/htbl/htbl_unsafe.ml
lyrm: Thanks ! It is actually an extra loop (compared to the original), that is needed because of the new type definition : `resize` can not call `assoc_node`. But I did not remember about this extra allocation!

test/htbl/stm_htbl.ml
lyrm: Good point, thanks ! I will do the changes for running only safe or unsafe version. For reducing the overall time of these tests, one way will be to avoid triggering resize in the dscheck tests.

bench/bench_htbl.ml
polytypic: I don’t think we need to change it in this case, because both hash table implementations should be essentially the same when it comes to inlining, but this is something to keep in mind when writing benchmarks.

bench/bench_htbl.ml
```diff
@@ -0,0 +1,83 @@
...
+        in
+        loop n
+    in
+    work ()
+  in
+
+  let config =
+    Printf.sprintf "%d workers, %d%% mem %d%% add %d%% rem" n_domains
+      percent_mem percent_add percent_rem
+  in
```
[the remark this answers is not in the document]
polytypic: Screenshot 2024-10-10 at 15 21 59

This coloring is due to not having unique configs.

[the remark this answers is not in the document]
polytypic: I'll look into adding a validation step into `multicore-bench` that would raise an error when this happens.  It shouldn't be too difficult.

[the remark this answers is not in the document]
polytypic: I have [a PR in multicore-bench](https://github.com/ocaml-multicore/multicore-bench/pull/29) where duplicates like this will raise a failure that I'll merge soon and publish a new version of multicore-bench.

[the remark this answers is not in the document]
polytypic: `multicore-bench` now checks that there are no duplicates and fails the bench if there are.

src_lockfree/htbl/dune
```diff
@@ -0,0 +1,15 @@
+(rule
+ (action
+  (with-stdout-to
+   htbl.ml
+   (progn
+    (cat htbl.head_safe.ml)
+    (cat htbl.body.ml)))))
```
polytypic, line 7:
```suggestion
    (echo "# 1 \"htbl.head_safe.ml\"\n")
    (cat htbl.head_safe.ml)
    (echo "# 1 \"htbl.body.ml\"\n")
    (cat htbl.body.ml)))))
```

src_lockfree/htbl/dune
```diff
@@ -0,0 +1,15 @@
...
+
+(rule
+ (action
+  (with-stdout-to
+   htbl_unsafe.ml
+   (progn
+    (cat htbl.head_unsafe.ml)
+    (cat htbl.body.ml)))))
```
polytypic, line 15:
```suggestion
    (echo "# 1 \"htbl.head_unsafe.ml\"\n")
    (cat htbl.head_unsafe.ml)
    (echo "# 1 \"htbl.body.ml\"\n")
    (cat htbl.body.ml)))))
```

src_lockfree/htbl/dune
polytypic: See ["Line number directives"](https://ocaml.org/manual/5.2/lex.html#sss:lex-linedir).

bench/bench_htbl.ml
```diff
@@ -0,0 +1,83 @@
...
+  assert (0 <= limit_mem && limit_mem <= 100);
+  assert (limit_mem <= limit_add && limit_add <= 100);
+
+  let t = Htbl.create ~hashed_type:(module Key) () in
+  if prepopulate then
+    for _ = 1 to n_keys do
+      let value = Random.bits () in
+      let key = value mod n_keys in
+      Htbl.try_add t key value |> ignore
+    done;
```
polytypic, line 27: This could be somewhat problematic.  This prepopulates the hash table.  However, the `work` function is run multiple times.  Given that the same random proportion of adds and removes is made the balance should not change much per run, but the balance might still drift over multiple runs.  So, for more consistent results it might make sense to remove everything from the hash table and prepopulate for each run.  (The benchmark in Picos does that.)

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+
+let[@tail_mod_cons] rec filter_ t msk chk = function
+  | Nil -> Nil
+  | Cons r ->
+      if t r.key land msk = chk then
+        Cons { r with rest = filter_ t msk chk r.rest }
+      else filter_ t msk chk r.rest
+
+let[@tail_mod_cons] rec filter_fst t mask chk = function
+  | Nil -> Nil_with_size { size_modifier = Size.used_once }
```
polytypic, line 153: `Nil_with_size { size_modifier = Size.used_once }` is logically a constant value.  However, because the `size_modifier` field is `mutable`, the compiler cannot optimize this to a constant value.  This means that a new value is allocated every time.  There are a couple of other places in the code where a similar `Nil_with_size { size_modifier = Size.used_once }` allocation happens unnecessarily.  Avoiding the allocations could slightly reduce memory use and improve performance.

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+        Cons_with_size
+          {
+            size_modifier = Size.new_once r.size Size.incr;
+            key;
+            value;
+            rest = Nil;
+          }
+      in
+      if Atomic_array.unsafe_compare_and_set r.buckets i (B before) (B after)
+      then (
```
polytypic, line 412: I'd prefer if we used `begin ... end` for blocks with semicolon separated statements.  I find them much easier to read and I like that the `begin ... end` alerts me that there are likely side-effects.  There are a few other places with `( ... )` blocks.

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+let to_seq t = snapshot t ~clear:false Backoff.default
+let remove_all t = snapshot t ~clear:true Backoff.default
+
+(* *)
+
+let rec try_find_random_non_empty_bucket size buckets seed i =
+  match Atomic_array.unsafe_fenceless_get buckets i with
+  | B (Nil_with_size nil_r) | B (Resize { spine = Nil_with_size nil_r }) ->
+      if nil_r.size_modifier != Size.used_once then
+        Size.update_once size nil_r.size_modifier;
```
polytypic, line 575: There are cases where it probably improves performance not to do so, but usually you want to clear the `size_modifier` field by assigning `Size.used_once` to it:

```suggestion
      if nil_r.size_modifier != Size.used_once then begin
        Size.update_once size nil_r.size_modifier;
        nil_r.size_modifier <- Size.used_once
      end;
```

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+let rec remove_node t key backoff =
+  let r = Atomic.get t in
+  let h = r.hash key in
+  let mask = Atomic_array.length r.buckets - 1 in
+  let i = h land mask in
+  match Atomic_array.unsafe_fenceless_get r.buckets i with
+  | B (Nil_with_size nil_r) ->
+      if nil_r.size_modifier != Size.used_once then (
+        Size.update_once r.size nil_r.size_modifier;
+        nil_r.size_modifier <- Size.used_once);
```
polytypic, line 463: Perhaps the clearing the `size_modifier` in this case might be best skipped.  The logic being that you clear it always after you introduce a new `size_modifier`.

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+                ('a, 'b, [ `Nil_with_size ]) tdt =
+              Nil_with_size { size_modifier }
+            in
+            if
+              Atomic_array.unsafe_compare_and_set r.buckets i (B before)
+                (B after)
+            then (
+              Size.update_once r.size nil_r.size_modifier;
+              nil_r.size_modifier <- Size.used_once;
+              adjust_size t r mask found_node)
```
polytypic, line 486: These three lines of code could probably be nicely folded into a single line:
```ocaml
adjust_size t r after mask found_node
```
The `adjust_size` function would then perform the `Size.update` on `after`.

Note that the native OCaml compiler can optimize the access of `size_modifier` such that no case discrimination (between `Nil_with_size` and `Cons_with_size`) is required as long as the field is at the same offset and there no other constructors are allowed.  Here is [an example on godbolt](https://ocaml.godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:12,fontUsePx:'0',j:1,lang:ocaml,selection:(endColumn:1,endLineNumber:12,positionColumn:1,positionLineNumber:12,selectionStartColumn:1,selectionStartLineNumber:12,startColumn:1,startLineNumber:12),source:'type+_+t+%3D%0A++%7C+Foo+:+%7B+mutable+x+:+int%3B+foo+:+bool+%7D+-%3E+%5B%3E+%60Foo%5D+t%0A++%7C+Bar+:+%7B+mutable+x+:+int%3B+bar+:+int+list+%7D+-%3E+%5B%3E+%60Bar%5D+t%0A++%7C+Extra_ball+:+%5B%3E+%60Extra_ball%5D+t%0A%0A(*+The+write+here+gets+translated+to+a+single+mov+--+no+pattern+matching+*)%0Alet+set+n+:+%5B%3C+%60Foo+%7C+%60Bar%5D+t+-%3E+_+%3D+function%0A++%7C+Foo+r+-%3E+r.x+%3C-+n%0A++%7C+Bar+r+-%3E+r.x+%3C-+n%0A%0A(*+The+read+here+gets+translated+to+a+single+mov+--+no+pattern+matching+*)%0Alet+get+:+%5B%3C+%60Foo+%7C+%60Bar%5D+t+-%3E+_+%3D+function%0A++%7C+Foo+r+-%3E+r.x+%0A++%7C+Bar+r+-%3E+r.x%0A'),l:'5',n:'1',o:'OCaml+source+%231',t:'0')),k:51.05977810895844,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:ocaml5200,filters:(b:'0',binary:'1',binaryObject:'1',commentOnly:'0',debugCalls:'1',demangle:'0',directives:'0',execute:'1',intel:'0',libraryCode:'0',trim:'1',verboseDemangling:'0'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:ocaml,libs:!(),options:'',overrides:!(),selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'+x86-64+ocamlopt+5.2.0+(Editor+%231)',t:'0')),header:(),k:48.940221891041574,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4) that shows that the native compiler indeed optimizes such code.

test/htbl/dscheck_htbl.ml
polytypic: > The point is to try multiple seeds to make it more probable to encounter a seed that actually triggers a resize as the choice of resizing is still partially made randomly.

This test is so expensive that I think we should just use a single seed that does trigger a resize and run the test once.  We could add a function to the hash table that returns the current number of buckets and use that to verify that a resize happened.

>  BTW, I was wondering if this (randomly deciding if resize is triggered) is still needed as the size computation is cheaper than before.

Actually it is more expensive than in the Picos version.  Both the Picos version and this Saturn version use a scalable counting mechanism.  They both allocate an atomic counter (or two) per domain.   What makes it expensive to read is that it requires reading the atomic counters of other domains.  This is expensive, because those counters are likely not in the cache of the current domain and so reading the counters potentially generates a lot of cache coherency traffic.  It also requires reading one (or two) counter(s) per domain.  The more domains you have, the more expensive it is to read the counter value.

src_lockfree/htbl/htbl.body.ml
```diff
@@ -0,0 +1,617 @@
...
+      rest
+  | Cons_with_size r -> begin
+      match rest with
+      | Nil_with_size r' ->
+          if r'.size_modifier != Size.used_once then
+            Size.update_once size r'.size_modifier;
+          Cons_with_size r
+      | Cons_with_size r' -> begin
+          if r'.size_modifier != Size.used_once then
+            Size.update_once size r'.size_modifier;
```
polytypic, line 230: We can actually perform an optimization here to reduce code size:
```suggestion
      begin
        match rest with
        | Nil_with_size r' ->
            if r'.size_modifier != Size.used_once then
              Size.update_once size r'.size_modifier;
        | Cons_with_size r' -> begin
            if r'.size_modifier != Size.used_once then
              Size.update_once size r'.size_modifier;
      end;
      match rest with
      | Nil_with_size r' ->
          Cons_with_size r
      | Cons_with_size r' -> begin
```
This may seem like it would generate more code, but the compiler can actually optimize the code inside the `begin ... end` such that the size update is done for either case with just a single copy of the code, because the `size_modifier` field is at the same offset in both cases and so no pattern matching is required.

src_lockfree/htbl/htbl.body.ml
lyrm: (took me a minute to get it, but got it, thanks !)

lyrm: So I defined a `nil_with_used_once` value to used in this case but it causes type issue with weak type variable. No sure if this is what you wanted me to do.

src_lockfree/htbl/htbl.body.ml
lyrm: I actually thought the reason to clear the size_modifier value was so further  call of `update_once` on it will be avoided, but I guess it is actually to ... clear memory ?

src_lockfree/htbl/htbl.body.ml
lyrm: I am not sure to get it. Is this just factorizing code by adding the size update on after in  `adjust_size` code  ?

src_lockfree/htbl/htbl.body.ml
polytypic: There are a few ways to avoid this problem:
* You could allow `Nil` at the root level of a bucket.
* You could include a `nil_with_used_once` as a field in the hash table record.
* You could use an unsafe cast to have a single `nil_with_used_once` value at the module level.

src_lockfree/htbl/htbl.body.ml
polytypic: It is a bit of both.

Note that operations that create new `Nil_with_size { size_modifier }` objects where `size_modifier` isn't initially `Size.used_once` call `Size.update_once` and overwrite the `size_modifier` with `Size.used_once`.  So, there is usually only a tiny window of time during which the `size_modifier` is not `Size.used_once`.  Performing `r.size_modifier <- Size.used_once` is actually rather expensive, because it calls `caml_modify`, so, while it makes sense to do it once, to free a bit of space and avoid repeated calls to `Size.update_once`, it also makes sense to avoid doing that mutation multiple times.

src_lockfree/htbl/htbl.body.ml
polytypic: Yes, the idea here is to reduce code size as the two lines (i.e. `Size.update_once ...; r.size_modifier <- ...` could be moved to happen in `adjust_size`.  If you look through the code, you will notice that every places that calls `adjust_size` also has such two lines.

The note with link to godbolt is to show that OCaml specifically does an optimization that makes this cheap.  The issue being that in some cases you have a `Nil_with_size` and in other case sa `Cons_with_size`.  Thanks to the optimization (and use of GADTs), you can write the code in such a way that the native OCaml compiler can completely avoid the pattern matching to distinguish between those two cases as the `size_modifier` field is at the same relative position in both cases.

polytypic:
BTW, I'm adding a couple of operations to the Picos Htbl:

```ocaml
val try_set : ('k, 'v) t -> 'k -> 'v -> bool
(** [try_set htbl key value] tries to update an existing binding of [key] to
    [value].  Returns [true] on success and [false] in case the hash table did
    not contain a binding for [key]. *)

val try_compare_and_set : ('k, 'v) t -> 'k -> 'v -> 'v -> bool
(** [try_compare_and_set htbl key before after] tries to update an existing
    binding of [key] from the [before] value to the [after] value.  The values
    are compared using physical equality, i.e. the [==] operator.  Returns
    [true] on success and [false] in case the hash table did not contain a
    binding of [key] to the [before] value. *)

val try_compare_and_remove : ('k, 'v) t -> 'k -> 'v -> bool
(** [try_compare_and_remove htbl key before] tries to remove a binding of [key]
    to the [before] value.  Values are compared using physical equality,
    i.e. the [==] operator.  Returns [true] on success and [false] in case the
    hash table did not contain a binding of [key] to the [before] value. *)
```

These are currently in a draft PR, [Add `Awaitable` atomic locations](https://github.com/ocaml-multicore/picos/pull/314), but I will likely move the Htbl changes to a separate PR soon.

There is a reason for adding operations with the above kind of semantics.  The reason being that those operations allow [consensus among arbitrarily many processes](https://dl.acm.org/doi/10.1145/62546.62593).  In practical terms, it means that you can create new non-blocking operations using those operations as building blocks.  If you look at the Stdlib Hashtbl, then many of the operations it provides, such as

```ocaml
val add : ('a, 'b) t -> 'a -> 'b -> unit
val remove : ('a, 'b) t -> 'a -> unit
val replace : ('a, 'b) t -> 'a -> 'b -> unit
```

do not allow consensus among many processes and are therefore limited in their use as building blocks of new non-blocking operations.

For example, in the PR, there is an (internal) lock-free operation

```ocaml
let rec add_awaiter t (one : Awaiters.is1) =
  match Htbl.find_exn awaiters (Packed t) with
  | before ->
      let many = Awaiters.snoc before one in
      if not (Htbl.try_compare_and_set awaiters (Packed t) before (Min1 many))
      then add_awaiter t one
  | exception Not_found ->
      if not (Htbl.try_add awaiters (Packed t) (Min1 one)) then
        add_awaiter t one
```

that adds an awaiter (`one` is a node for single awaiter) to a queue in the `awaiters` hash table.  Notice how the operations provided by the `Htbl` allow one to write that using a very familiar lock-free retry pattern.

lyrm:
This PR is ready. I have began squashing the commits to make it more readable if additional people want to have a look before it is merged.

PR #155, opened 2024-10-08
Blocking bounded stack

lyrm:
Add  a bounded stack based on Treiber stack implementation. 

Todo: 

- [x] Documentation
- [ ] Tests for pop/push/peek
- [x] DScheck tests
- [x] Benchmarks

### DIFF
--- a/bench/bench_bounded_stack.ml
+++ b/bench/bench_bounded_stack.ml
@@ -1,0 +1,72 @@
+open Multicore_bench
+module Stack = Saturn_lockfree.Bounded_stack
+
+let run_one_domain ~budgetf ?(n_msgs = 50 * Util.iter_factor) () =
+  let t = Stack.create () in
+
+  let op push = if push then Stack.push t 101 else Stack.pop_opt t |> ignore in
+
+  let init _ =
+    assert (Stack.is_empty t);
+    Util.generate_push_and_pop_sequence n_msgs
+  in
+  let work _ bits = Util.Bits.iter op bits in
+
+  Times.record ~budgetf ~n_domains:1 ~init ~work ()
+  |> Times.to_thruput_metrics ~n:n_msgs ~singular:"message" ~config:"one domain"
+
+let run_one ~budgetf ?(n_adders = 2) ?(n_takers = 2)
+    ?(n_msgs = 50 * Util.iter_factor) () =
+  let n_domains = n_adders + n_takers in
+
+  let t = Stack.create () in
+
+  let n_msgs_to_take = Atomic.make 0 |> Multicore_magic.copy_as_padded in
+  let n_msgs_to_add = Atomic.make 0 |> Multicore_magic.copy_as_padded in
+
+  let init _ =
+    assert (Stack.is_empty t);
+    Atomic.set n_msgs_to_take n_msgs;
+    Atomic.set n_msgs_to_add n_msgs
+  in
+  let work i () =
+    if i < n_adders then
+      let rec work () =
+        let n = Util.alloc n_msgs_to_add in
+        if 0 < n then begin
+          for i = 1 to n do
+            Stack.push t i
+          done;
+          work ()
+        end
+      in
+      work ()
+    else
+      let rec work () =
+        let n = Util.alloc n_msgs_to_take in
+        if n <> 0 then begin
+          for _ = 1 to n do
+            while Option.is_none (Stack.pop_opt t) do
+              Domain.cpu_relax ()
+            done
+          done;
+          work ()
+        end
+      in
+      work ()
+  in
+
+  let config =
+    let format role n =
+      Printf.sprintf "%d %s%s" n role (if n = 1 then "" else "s")
+    in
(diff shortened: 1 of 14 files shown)


lyrm:
The benchmarks are currently exactly the same as the treiber stack ones, i.e. it uses `push` and `pop_opt` with a stack with a capacity of `Int_max`. I found it useful this way because it enables to see the overhead added by both the capacity field and the recurring calls to `signal_all`. 

I don't see how benchmarking functions that require a Picos' scheduler will be useful as the scheduler itself will have an effect on it. 

What do you think @polytypic ?

src_lockfree/bounded_stack.mli
```diff
@@ -0,0 +1,113 @@
...
+
+val pop_all : 'a t -> 'a list
+(** [pop_all stack] removes and returns all elements of the [stack] in the reverse 
+order they were pushed. *)
+
+(** {2 Producer functions} *)
+
+val push : 'a t -> 'a -> unit
+(** [push stack element] adds [element] to the top of the [stack].
+    Raises an exception if the stack is full. *)
```
polytypic, line 45: Looking at the implementation this seems to block rather than raise an exception in case the stack is full.

src_lockfree/bounded_stack.mli
```diff
@@ -0,0 +1,113 @@
...
+      # Saturn.Bounded_stack.push t 1
+      - : unit = ()
+      # Saturn.Bounded_stack.pop t
+      - : int = 1
+      # Saturn.Bounded_stack.peek_opt t
+      - : int option = Some 42
+      # Saturn.Bounded_stack.pop_opt t
+      - : int option = Some 42 
+      # Saturn.Bounded_stack.pop_opt t
+      - : int option = None ]}
```
polytypic, line 69: We should have these checked by MDX.

bench/bench_bounded_stack.ml
```diff
@@ -0,0 +1,72 @@
...
+  |> Times.to_thruput_metrics ~n:n_msgs ~singular:"message" ~config:"one domain"
+
+let run_one ~budgetf ?(n_adders = 2) ?(n_takers = 2)
+    ?(n_msgs = 50 * Util.iter_factor) () =
+  let n_domains = n_adders + n_takers in
+
+  let t = Stack.create () in
+
+  let n_msgs_to_take = Atomic.make 0 |> Multicore_magic.copy_as_padded in
+  let n_msgs_to_add = Atomic.make 0 |> Multicore_magic.copy_as_padded in
```
polytypic, line 25: Multicore-bench now has a [`Countdown` abstraction](https://ocaml-multicore.github.io/multicore-bench/doc/multicore-bench/Multicore_bench/Countdown/index.html) for this.

src_lockfree/bounded_stack.ml
```diff
@@ -0,0 +1,156 @@
...
+  let prod_waiters = Atomic.make_contended [] in
+  { head; cons_waiters; prod_waiters; capacity = max capacity 1 }
+  |> Multicore_magic.copy_as_padded
+
+let is_empty t = Atomic.get t.head == Nil
+
+let rec signal_all waiters =
+  let triggers = Atomic.get waiters in
+  if triggers != [] then
+    if Atomic.compare_and_set waiters triggers [] then
```
polytypic, line 24: You can use `Atomic.exchange` here and avoid the retry loop.

src_lockfree/bounded_stack.ml
```diff
@@ -0,0 +1,156 @@
...
+      List.iter Trigger.signal triggers
```
polytypic, line 25: Note that this signals the triggers in LIFO order.  Ideally we might want to signal triggers in FIFO order, to hopefully improve fairness, and try to avoid signaling all of them, to avoid the [thundering herd](https://en.wikipedia.org/wiki/Thundering_herd_problem) phenomena.  At this point I mention these mostly as things to work on later.

src_lockfree/bounded_stack.ml
```diff
@@ -0,0 +1,156 @@
...
+  prod_waiters : Trigger.t list Atomic.t;
+  capacity : int;
+}
+
+let create ?(capacity = Int.max_int) () =
+  let head = Atomic.make_contended Nil in
+  let cons_waiters = Atomic.make_contended [] in
+  let prod_waiters = Atomic.make_contended [] in
+  { head; cons_waiters; prod_waiters; capacity = max capacity 1 }
+  |> Multicore_magic.copy_as_padded
```
polytypic, line 17: Note that all the padding here takes quite a bit of memory.  A single non-padded atomic takes only 2 words.  Here we have 5 padded objects.  Each padded object takes 8 or 16 words (possibly even 32 on some architectures).  So, this allocates 40 or 80 words or 320 or 640 bytes, which is, relatively speaking, a lot.  At this point I'm mostly mentioning this as something to work on later.

lyrm:
Closed for now (see #156).

PR #156, opened 2024-10-28
Bounded stack

lyrm:
Based on #155. This PR removes the blocking part of the previous PR. The issue with it is that it adds a dependency to `picos`, probably requiring to create a `Saturn_picos` package.

For Saturn.1.0,  we have decided to do only a bounded stack, thus this PR.

### DIFF
--- a/src/bounded_stack.ml
+++ b/src/bounded_stack.ml
@@ -99,20 +99,6 @@ let rec push_as : type r. 'a t -> Backoff.t -> 'a -> r mono -> r =
 let push_exn t value = push_as t Backoff.default value Unit
 let try_push t value = push_as t Backoff.default value Bool
 
-type ('a, _) poly3 = Value : ('a, 'a) poly3 | Bool : ('a, bool) poly3
-
-let rec set_as : type v r. v t -> v -> Backoff.t -> (v, r) poly3 -> r =
- fun t value backoff poly ->
-  match Atomic.get t.head with
-  | _, [] -> ( match poly with Value -> raise Empty | Bool -> false)
-  | (len, hd :: tl) as old_head ->
-      if Atomic.compare_and_set t.head old_head @@ (len, value :: tl) then
-        match poly with Value -> hd | Bool -> true
-      else set_as t value (Backoff.once backoff) poly
-
-let set_exn t value = set_as t value Backoff.default Value
-let try_set t value = set_as t value Backoff.default Bool
-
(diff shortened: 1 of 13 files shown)


lyrm:
@Sudha247 : Could you have a look at this PR (at the documentation, or even a quick look at the tests, maybe ?), any feedbacks will be great !

src/bounded_stack.ml
```diff
@@ -0,0 +1,208 @@
+type 'a node = Nil | Cons of { value : 'a; tail : 'a node; capacity : int }
+type 'a t = { head : 'a node Atomic.t; capacity : int }
```
art-w, line 2: The `Cons capacity` field looks a bit redundant since it's only really useful near the root instead of at every level of the node list. Maybe `type 'a t = (int * 'a list) Atomic.t` would be easier? (with the `int` representing the room left in the stack, so when it reaches 0 we know it's full)

```ocaml
let capacity t = (* not needed, just to clarify the semantics *)
  let room_left, lst = Atomic.get t in room_left + List.length lst
```

But the O(1) `length` function was nice to have, so perhaps:

```ocaml
type 'a t = { head: (int * 'a list) Atomic.t ; capacity: int }
let length t = t.capacity - fst (Atomic.get t.head)
let is_full t = fst (Atomic.get t.head) = 0

(* or the fst could represent the list length: *)
let length t = fst (Atomic.get t.head)
let is_full t = t.capacity = length t
```

On a side-note, the `is_full` function could be nice to add to the API? :)

lyrm: Good point, thanks !

lyrm:
@art-w I think this PR is good to. What do you think ?

art-w (approved):
Yes it looks good! Out of curiosity, are the `copy_as_padded` on records necessary? (my understanding is that this makes the allocated object fatter, so as to avoid false-sharing contention, but the `head` atomic already appears to be protected by `make_contended`)

lyrm:
Mmh there is a stack overflow on `Treiber_stack.push_all`. This PR has almost the same implementation so I will have a look before merging !

lyrm:
> Yes it looks good! Out of curiosity, are the `copy_as_padded` on records necessary? (my understanding is that this makes the allocated object fatter, so as to avoid false-sharing contention, but the `head` atomic already appears to be protected by `make_contended`)

Good question. As nothing else is mutable in the stack type, probably not.

PR #158, opened 2024-10-31
Add new functions and improve documentations of Treiber stack.

lyrm:
Add functions to the Treiber stack, to match bounded stack in #156 : 
- With list : `of_list`, `push_all`, `pop_all` 
- With seq: `of_seq`, `add_seq`, `to_seq` 
- Consumer functions : `peek_exn`, `peek_opt`, `pop_exn`, `drop_exn`
- Set functions : `set_exn`, `try_exn`
- Others : `try_compare_and_pop`, `try_compare_and_set`

### DIFF
--- a/src/treiber_stack.mli
+++ b/src/treiber_stack.mli
@@ -47,13 +47,6 @@ val drop_exn : 'a t -> unit
 
   @raises Empty if the [stack] is empty. *)
 
-val try_compare_and_pop : 'a t -> 'a -> bool
-(** [try_compare_and_pop stack before] tries to remove the top element of the 
-  [stack] if it is equal to [before]. Returns [true] on success and [false] in 
-  case the stack is empty or if the top element is not equal to [before].
-
-  ℹ️ The values are compared using physical equality, i.e. the [==] operator. *)
-
 val pop_all : 'a t -> 'a list
 (** [pop_all stack] removes and returns all elements of the [stack] in LIFO 
 order. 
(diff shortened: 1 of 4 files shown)


src/treiber_stack.mli
```diff
@@ -1,31 +1,191 @@
...
-(** Raised when {!pop} is applied to an empty queue. *)
+(** Raised when {!pop_exn}, {!peek_exn}, {!drop_exn}, or {!set_exn} is 
+  applied to an empty stack.
 
-val pop : 'a t -> 'a
-(** [pop s] removes and returns the topmost element in the
-    stack [s].
+  This exception is meant to avoid allocations required by an option type.
+  As such, it does not register backtrace information and it is recommended to 
+  use the following pattern to catch it.
```
art-w, line 28: Do we have examples of other libraries which do this? My understanding from [OCaml behind the scenes: Exception (around 47m12s)](https://watch.ocaml.org/w/vur2uhCFTZWQfDfkcbCPfC?start=47m12s) is that `raise Empty` is already allocation-free unless the developer has explicitly requested debugging backtraces (cc @fabbing in case I'm wrong ^^')

If maximum performances are required, would the following very verbose optin alternative API be allocation-free? (or would the cost of calling `default ()` be prohibitive?)

```ocaml
match pop_with s ~default:(fun () -> raise_notrace Empty) with
| value -> ...
| exception Empty -> ...
```

art-w (commented):
This looks very nice! I had some minor questions, but otherwise I'm curious if you have examples use-cases for `try_set`, `try_compare_and_set` and `set_exn`? :)

lyrm:
For the new functions, `try_set`, `try_compare_and_set`, and `set_exn`, I was considering whether they could be useful for some form of message passing.

Originally, these functions were introduced for the hashtable from the picos library (see picos [PR 317](https://github.com/ocaml-multicore/picos/pull/317)). In the context of the hashtable, these operations make a lot of sense. However, due to time constraints, I haven’t had the chance to fully explore how these could be beneficial for a stack. I might move this part of the PR to a subsequent draft PR instead. What do you think ?

art-w:
Yes I've plenty of use cases for those functions with the hashtable, but I can't think of any for the stack... so it makes sense to remove them for the time being and reconsider adding them later once a use case shows up :)

lyrm:
@art-w : Thanks a lot for your reviews ! Anything else before I merge this ?

art-w (approved):
LGTM! :tada:

PR #159, opened 2024-11-04
Remove Saturn_lockfree package.

lyrm:
Currently, `Saturn` includes only lock-free data structures, with no blocking structures implemented. In the future, adding blocking data structures would likely introduce dependencies on either `picos` or a specific scheduler. This would suggest creating a `Saturn_picos` library (or `Saturn_[scheduler_name]` as appropriate). 

Consequently, having a `Saturn_lockfree` package seems to no longer make sense. Thus, this PR that simply removes it !


lyrm:
Note: the README will be updated in a following PR. 

@polytypic or @Sudha247 : can one of you do a quick review ?

Sudha247 (approved):
LGTM!

Unrelated to this PR, it would be good to write down the plans for blocking structures.

PR #162, opened 2024-11-19
Mpsc update

lyrm:
The `mpsc` queue has been updated to align with the code in `Eio`. Previously, it was based on an `mpmc` (multi-producer, multi-consumer) queue, but the new implementation leverages the fact that there is only one consumer. 

**Important Note**: This change limits the usage of the `is_empty` and `close` functions, which can now only be called by the consumer.

### Changes Made:
- Updated the implementation to match the new `Eio` version.
- Removed non-functional QCheck tests (`close` can not be called by a producer).
- Added a `drop_exn` function
- Added `push_all` and `of_list` functions
- Renamed `pop` and `peek`  `pop_exn` and `peek_exn` respectively
- Updated STM tests to reflect the new behavior.
- Revised the documentation to include the new limitations and features. 

Note: use of a `backoff` seems to not improve performance.

### DIFF
--- a/bench/bench_mpsc.ml
+++ b/bench/bench_mpsc.ml
@@ -1,0 +1,79 @@
+open Multicore_bench
+module Queue = Saturn.Single_consumer_queue
+
+let run_one_domain ~budgetf ?(n_msgs = 50 * Util.iter_factor) () =
+  let t = Queue.create () in
+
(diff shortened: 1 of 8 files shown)


lyrm:
@art-w : this is ready to review.

art-w (approved):
LGTM!

PR #185, opened 2025-05-05
Bench queues and stacks with heap allocated blocks

polytypic:
Queues and stacks were previously benchmarked with immediate values only, which unfortunately partially hides the potential cost of write barriers related to setting and clearing elements.

**_Partially hiding the cost of write barriers makes the queue benchmarks unrealistic._**  Queues and stack are very rarely used with only immediate values.

### DIFF
--- a/bench/bench_bounded_queue.ml
+++ b/bench/bench_bounded_queue.ml
@@ -9,7 +9,7 @@ module Make (Bounded_queue : Bounded_queue_intf.BOUNDED_QUEUE) : BENCH = struct
     let t = Bounded_queue.create () in
 
     let op push =
-      if push then Bounded_queue.try_push t 101 |> ignore
+      if push then Bounded_queue.try_push t (ref push) |> ignore
       else Bounded_queue.pop_opt t |> ignore
     in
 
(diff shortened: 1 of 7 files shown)


lyrm:
Thanks for the PR. I am reviewing it today. Also, I am working on fixing the CI issue.

lyrm (approved):
I’ve fixed the CI issue. Could you rebase to check if it’s working on your side as well?

Thanks again for the PR! Have you run any benchmarks to see if this changes how the different queues compare to each other ? What about the two-stack queue for example ?

polytypic:
> Have you run any benchmarks to see if this changes how the different queues compare to each other ?

I changed the benchmarks in [multicore-bench](https://bench.ci.dev/ocaml-multicore/multicore-bench/branch/main?worker=pascal&image=bench.Dockerfile) and [picos](https://bench.ci.dev/ocaml-multicore/picos/branch/main?worker=pascal&image=bench.Dockerfile) to allocate the elements just like in this PR.

> What about the two-stack queue for example ?

The two-stack queue doesn't need to clear/set values, which means that allocating the elements only adds the allocation cost.

